PR #5, opened 2024-11-15
fix: Prevent panic in acp policy-ids query

iverc:
## Description

This PR adds a fix to the `PolicyIds` keeper in the ACP module to prevent panics when querying policy-ids with no policies created.

## Tasks
- [x] Add `MapNullableSlice` function
- [x] Use `MapNullableSlice` instead of `MapSlice` in `PolicyIds` keeper

### DIFF
--- a/utils/functional.go
+++ b/utils/functional.go
@@ -29,3 +29,17 @@ func MapFailableSlice[T any, U any](ts []T, mapper func(T) (U, error)) ([]U, err
 
 	return us, nil
 }
+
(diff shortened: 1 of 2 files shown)


jsimnz (approved):

jsimnz:
All looks good, straight forward 👍

Lodek (approved):
Looks good but I'm confused, did you catch a panic doing this? The ListPolicy method shouldn't return any nil pointers. If you have a panic trace please let me know and I'll apply the fix on acp_core to assure that doesn't happen. PR is good though, feel free to mege it

PR #8, opened 2024-12-02
fix: Tier module fixes and improvements (part 2)

iverc:
## Description

This PR aims to add a number of improvements and features to the Tier module, as well as to fix multiple bugs in the existing logic.

## Tasks
- [x] Implement `CancelUnlocking` keeper and msg server.
- [x] Implement `LockupKeyToAddressesAtHeight` and `UnlockingLockupKey` to store unlocking lockups as `delAddr/valAddr/creationHeight/`.
- [x] Implement `MustIterateUnlockingLockups` to iterate unlocking lockups.
- [x] Add `GetLockupAmount` and `removeLockup`.
- [x] Fix issues with the existing `Unlock` logic and use `TokensFromSharesTruncated` for the unlocking lockup amount calculation to prevent discrepancies in shares calculations.
- [x] Fix issues with the existing `CompleteUnlocking` logic.
- [x] Fix issues with the existing `GetAllLockups` logic.
- [x] Add distribution keeper to the Tier module and expected keepers.
- [x] Add `creation_height` field to `UnlockingLockupRequest`.
- [x] Improve the test `TierKeeper` implementation.
- [x] Add tests for `Lock`, `Unlock`, `Redelegate`, `CompleteUnlocking`, `CancelUnlocking`, and `MintCredit`.
- [x] Add mocks for all expected keepers. 
- [x] Add mock tests for `Lock`, `Unlock`, and `Redelegate`.
- [x] Add `CreateTestValidator` and `CreateTestEncodingConfig` utils for testing.
- [x] Add tests for `MustIterateUnlockingLockups` and `IterateLockups`.
- [x] Document changes to `RewardRates` *Rate* param to make it clear that we use integers to represent percentages with 2 decimal precision (e.g. a rate of 150 represents 1.50).

Changes to be included in the upcoming PR:
- Prorate credit logic improvements and related tests
- Events for all tier msgs
- E2e tests for the lock, unlock, complete/cancel unlocking flow
- Changes related to custom rewards distribution and sliding scale APY
- Slashing logic improvements for the developer stake and related tests


testutil/keeper/tier.go
jsimnz, line 1: praise: great work expanding out the testutils for this 👏

x/tier/keeper/credit.go
```diff
@@ -117,10 +126,10 @@ func (k Keeper) resetAllCredits(ctx context.Context) error {
 	return nil
 }
 
-// calculateCredit calculates the reward earned on the lockingAmt.
+// CalculateCredit calculates the reward earned on the lockingAmt.
 // lockingAmt is stacked up on top of the lockedAmt to earn at the
 // highest eligible reward.
-func calculateCredit(rateList []types.Rate, lockedAmt, lockingAmt math.Int) math.Int {
+func CalculateCredit(rateList []types.Rate, lockedAmt, lockingAmt math.Int) math.Int {
```
jsimnz: question: Is there a specific reason to make this a public API?

x/tier/keeper/lockup.go
```diff
@@ -19,33 +19,73 @@ import (
...
+	k.MustIterateUnlockingLockups(ctx, unlockingLockupsCallback)
+	k.MustIterateLockups(ctx, lockupsCallback)
 
 	return lockups
 }
 
-func (k Keeper) SetLockup(ctx context.Context, unlocking bool, delAddr sdk.AccAddress, valAddr sdk.ValAddress, amt math.Int, creationHeight int64, unbondTime *time.Time, unlockTime *time.Time) {
+// SetLockup stores or updates a lockup in the state based on the key from LockupKey/UnlockingLockupKey.
+// We normalize lockup times to UTC before saving to the store for consistentcy.
+func (k Keeper) SetLockup(ctx context.Context, unlocking bool, delAddr sdk.AccAddress, valAddr sdk.ValAddress, amt math.Int,
```
jsimnz: question: The added `unbondTime` and `unlockTime` as parameters is interesting. Should this be a module parameter compared to an argument on a public method for the keeper? What instances would there be for using non-standard (computed based on context by the `SetLockup` function itself). My primary concern is that its a larger surface area to accidentally screw up in the future.

But, perhaps there is a notable/important usecase I'm missing that makes this necessary.

x/tier/keeper/lockup.go
jsimnz, line 1: question: I might have missed it, but is there still an API for getting all the lockups for a specific delegator? Instead of needing to know the delegator/validator pair ahead of time?

x/tier/keeper/lockup.go
jsimnz: Same question for `creationHeight` (I know this was in the API signature before your changes). Should this be a computed argument?

x/tier/types/events.go
```diff
@@ -0,0 +1,5 @@
+package types
+
+const (
+	EventTypeCancelUnlocking = "cancel_unlocking"
```
jsimnz: suggestion: Nothing todo with this exact line, just commenting here since it provides context.

We should probably map all the various/necessary/potential event types that would be of value for the `tier` module and enumerate them all here.

jsimnz (approved):
Really great stuff, especially the testing work added here. 

Have a few main questions:
- exposing the `CalculateCredit` function
- some lockup API arguments

But beyond that looks great! Approving now so we can quickly go through the answers and get to consensus, without you needing to wait for me to come back and finally approve, just don't merge it yet 👍 .

x/tier/keeper/lockup.go
iverc: Currently there is no API for getting all lockups of a specific delegator.

Added `GetLockups` and related test in `iverc/tier-module-improvements-3`.

x/tier/types/events.go
iverc: Agree. Other events are defined in `iverc/tier-module-improvements-3` (https://github.com/sourcenetwork/sourcehub/commit/13504e50f06e0e4c741290ac38fdfb8bdb07a85e)

x/tier/keeper/lockup.go
iverc: Thank you for pointing it out! I agree with your concerns - this implementation is indeed not ideal. The main reason why I added `unbondTime` and `unlockTime` initially was to be able to use it in `InitGenesis` to load existing lockups with correct height and time(s).

I made required changes to compute`creationHeight`, `unlockTime`, and `unbondTime` within `SetLockup`, and added `SaveLockup` to be used in the `InitGenesis` instead (in `iverc/tier-module-improvements-3` branch).

x/tier/keeper/credit.go
iverc: Nice catch! Initially I made this change to enable it in tests with the `keeper_test` package. 

Refactored `calculateCredit` and related tests into a separate file with the `keeper` package to keep it private and avoid import cycles (https://github.com/sourcenetwork/sourcehub/commit/afdb7f899add6b6d5ba816756cee9523d479b279)

PR #13, opened 2025-01-20
fix: Tier module fixes and improvements (part 4)

iverc:
## Description
This PR aims to add a number of fixes and additional improvements on top of work done in https://github.com/sourcenetwork/sourcehub/pull/10

## Tasks
- [x] Add `UnlockingLockup`, refactor `Lockup`, `GenesisState`, and query messages.
- [x] Rename `unbondTime` to `completionTime` to match native staking module.
- [x] Add `creation_height` to  `MsgUnlockResponse`.
- [x] Store lockups / unlocking lockups separately in `GenesisState`.
- [x] Fix timestamp and calculation issues in `calculateProratedCredit`.
- [x] Improve logic related to storing/retrieving lockups and unlocking lockups. 
- [x] Handle `completionTime`/`unlockTime` edge case in `CompleteUnlocking`.
- [x] Refactor `proratedCredit`, add `TestCalculateProratedCredit` and `TestProratedCredit`.
- [x] Add multiple test suites, update/fix a bunch of tests.

PR #15, opened 2025-01-29
fix: Tier module fixes and improvements (part 5)

iverc:
## Description
This PR adds additional fixes and improvements, following up work done in https://github.com/sourcenetwork/sourcehub/pull/13

## Tasks
- [x] Use `$OPEN` as the default denom in all modules / params
- [x] Fix `unlocking-lockup` cli command and rpc endpoint
- [x] Set default mint/staking/crisis/gov module params in the config


app/inflation.go
```diff
@@ -0,0 +1,35 @@
...
+	stakingKeeper *stakingkeeper.Keeper,
+) minttypes.InflationCalculationFn {
+	// Adjust bonded ratio based on the x/tier module developer stake so that it does not affect rewards
+	return func(ctx context.Context, minter minttypes.Minter, params minttypes.Params, bondedRatio math.LegacyDec) math.LegacyDec {
+		devStake := tierKeeper.GetDeveloperStake(sdk.UnwrapSDKContext(ctx))
+		totalSupply := bankKeeper.GetSupply(ctx, appparams.DefaultBondDenom).Amount
+
+		totalBonded, err := stakingKeeper.TotalBondedTokens(ctx)
+		if err != nil {
+			totalBonded = math.ZeroInt()
```
Lodek: Do we really want to handle this error silently? It seems staking keeper only fails if it can fetch the denom https://github.com/cosmos/cosmos-sdk/blob/v0.50.11/x/staking/keeper/pool.go#L81. Maybe a panic because otherwise `adjustedBonded` and `adjustedBondedRation` would be negative (not sure if that's a problem).
If this is a safe default, a comment explaining why would be helpful.

Lodek (approved):
LGTM. My only suggestion would be to extract the Inflation function as a pure function and maybe add a simple test to it? It seems important enough to warrant that.

app/inflation.go
jsimnz: I agree with the panic here, perhaps we can even add a check at the top of the `ProvideInflationCalculationFn` to run the same check to ensure that the denomination exists. That way we can panic during the `ProvideInflationCalculationFn` which happens at startup.

Unless theres a legit reason why the denom might not exist at startup, but I cant think of any.

app/inflation.go
```diff
@@ -0,0 +1,35 @@
...
+	"cosmossdk.io/math"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
+	minttypes "github.com/cosmos/cosmos-sdk/x/mint/types"
+	stakingkeeper "github.com/cosmos/cosmos-sdk/x/staking/keeper"
+	appparams "github.com/sourcenetwork/sourcehub/app/params"
+	tierkeeper "github.com/sourcenetwork/sourcehub/x/tier/keeper"
+)
+
+func ProvideInflationCalculationFn(
```
jsimnz: question: Are we sure that this custom inflation function is actually necessary?

The tokenomics basically wants to compute the inflation based on total supply vs total locked (delegator stake + dev lock). 

image

What we want to pass into the `minter.NextInflationFunction` at the end of this function is $R_n$ which is defined as $(stake_{delegator} + lock_{dev}) / total_{supply}$

My understanding of the custom function is that you've done is remove the $lock_{dev}$ from this calculation. Which would *increase* the inflation instead of what we want which is effectively *decreasing* inflation.

app/inflation.go
iverc: Thanks, addressed it in https://github.com/sourcenetwork/sourcehub/pull/20

jsimnz (changes requested):
Issue regarding the custom inflation function (from my POV).

This implementation would only work if the dev stake wasn't already being accounted for in the total stake, but since its currently implemented as a regular delegation owned by the tier module the default inflation function is correct, with the exception that we need to burn the rewards minted to the tier module. 

At least this is my understanding based on things, maybe im wrong 🙃

app/inflation.go
jsimnz: I think the technically correctly updates the inflation ratio to be smaller as intended, but the problem is that the rewards aren't calculated correctly down stream, since the `tier` module account still gets minted rewards.

So we have correctly reduced the rewards, but now we distributed a smaller amount of rewards to the same number of delegated tokens.

iverc: As discussed, removed the custom inflation function.

jsimnz (approved):
LGTM.

PR #17, opened 2025-01-30
fix: App and configuration fixes

iverc:
## Description
This PR adds multiple fixes and improvements to the core app logic and related configs, following up work done in https://github.com/sourcenetwork/sourcehub/pull/15

## Tasks
- [x] Add ante handler with basic decorators.
- [x] Use `uopen` as the base denom and fix incorrect exponents in denoms metadata.
- [x] Use default overrides for `bank`, `crisis`, `gov`, `mint`, and `staking` module basics.
- [x] Update app/params constants and use them instead of hardcoded strings.
- [x] Set default `BaseDenom`/ `MinGasPrices` in e2e test configs to fix failing tests.
- [x] Set default `MinGasPrices` in the cmd server config.
- [x] Update genesis script and ignite configs to use correct denom and values.


app/app.go
```diff
@@ -146,31 +147,21 @@ func init() {
...
-	return govProposalHandlers
-}
-
 // AppConfig returns the default app config.
 func AppConfig() depinject.Config {
 	return depinject.Configs(
 		appConfig,
-		// Loads the ao config from a YAML file.
+		// Loads the app config from a YAML file.
 		// appconfig.LoadYAML(AppConfigYAML),
```
Lodek: since the yaml loading is commented out, can we remove this block?

app/app.go
```diff
@@ -146,31 +147,21 @@ func init() {
...
 		depinject.Supply(
 			// supply custom module basics
 			map[string]module.AppModuleBasic{
 				genutiltypes.ModuleName: genutil.NewAppModuleBasic(genutiltypes.DefaultMessageValidator),
-				govtypes.ModuleName:     gov.NewAppModuleBasic(getGovProposalHandlers()),
-				// this line is used by starport scaffolding # stargate/appConfig/moduleBasic
```
Lodek, line 173: question: will this break ignite if we need to add another submodule in the future?

Lodek (approved):
lgtm, good job

app/app.go
iverc: it won't break ignite, we'll just need to set module basics manually

Dockerfile
```diff
@@ -37,7 +37,7 @@ RUN --mount=type=cache,target=/root/.cache make build
 
 # Dev image entrypoint
 ENTRYPOINT ["scripts/dev-entrypoint.sh"]
-CMD ["start", "--minimum-gas-prices=0.001uopen"]
+CMD ["start"]
```
jsimnz, line 40: question: By removing this parameter (which I thought was required with the current version of the cosmos sdk) mean that calling this Dockerfile after its built, you must specify this parameter manually?

app/ante/ante.go
```diff
@@ -0,0 +1,51 @@
...
+import (
+	"cosmossdk.io/x/tx/signing"
+	sdk "github.com/cosmos/cosmos-sdk/types"
+	"github.com/cosmos/cosmos-sdk/x/auth/ante"
+	bankkeeper "github.com/cosmos/cosmos-sdk/x/bank/keeper"
+	ibcante "github.com/cosmos/ibc-go/v8/modules/core/ante"
+	ibckeeper "github.com/cosmos/ibc-go/v8/modules/core/keeper"
+)
+
+func NewAnteHandler(
```
jsimnz, line 13: todo: Add a quick function comment that describes why this manually defined set of AnteHandlers is necessary (It was my understanding that many of these were set by default, but perhaps I am wrong here).

Dockerfile
jsimnz: ahh, nvm. I see youre setting the default gas via code here `srvCfg.MinGasPrices = fmt.Sprintf("%v%s", appparams.DefaultMinGasPrice, appparams.DefaultBondDenom)`

scripts/ignite/config-tier.yml
jsimnz, line 1: todo: Seems to be a missed merge conflict syntax in this file😅

jsimnz (changes requested):
Looks really great. Especially like the `overrides` package to customize the genesis config for each module.

Everything is good to go for the most part. One or two minor things.

However, there is a file with some left over merge confict git syntax that I marked.

scripts/ignite/config-tier.yml
iverc: nice catch, thank you!

app/ante/ante.go
iverc: Thanks, added a comment. We need it to extend the default ante handler with additional/custom decorators (e.g. `NewRedundantRelayDecorator`)

jsimnz: Can you separate which handlers are part of the 'default' flow and which are added/custom. 

Or just comment which in case it's order sensitive

jsimnz (approved):
Lgtm, added an additional request to the ante handler thread.

Approving now though. Appreciate it!

PR #19, opened 2025-02-04
fix: Acp module fixes and improvements (part 1) 

iverc:
## Description
This PR adds multiple fixes and improvements to the ACP module.

## Tasks
- [x] Use `Querier` for all acp queries.
- [x] Fix `acp/keeper/main_test.go` to make sure that we run acp keeper tests.
- [x] Add more tests for acp queries.


Lodek (approved):
lgtm, in the future these tests should go into the integration suite instead, that way we can test for all "frontends", but we don't have to worry about that until we sort out a better e2e test system

Lodek:
Also, feel free to merge this PR before mine, I can rebase on top of it and move the new queries to the querier

PR #20, opened 2025-02-09
fix: Improve custom inflation function

iverc:
## Description

This PR adds changes to improve the inflation function performance, handle edge cases, and some minor fixes.

## Tasks
- [x] Add `GetTotalLockupsAmount` and `SetTotalLockupsAmount` methods.
- [x] Track total lockups amount in `AddLockup` and `SubtractLockup`. 
- [x] Use `GetTotalLockupsAmount` in the custom inflation function to avoid iterating all records every block.
- [x] Handle multiple edge cases during inflation calculations. 
- [x] Add new tests and update existing tests.

### DIFF
--- a/x/tier/keeper/lockup.go
+++ b/x/tier/keeper/lockup.go
@@ -165,12 +165,22 @@ func (k Keeper) removeUnlockingLockup(ctx context.Context, delAddr sdk.AccAddres
 }
 
 // AddLockup adds provided amt to the existing delAddr/valAddr lockup.
-func (k Keeper) AddLockup(ctx context.Context, delAddr sdk.AccAddress, valAddr sdk.ValAddress, amt math.Int) {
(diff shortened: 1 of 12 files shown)


x/tier/keeper/lockup.go
```diff
@@ -340,3 +361,32 @@ func (k Keeper) lockupStore(ctx context.Context, unlocking bool) prefix.Store {
...
+func (k Keeper) GetTotalLockupsAmount(ctx context.Context) (total math.Int) {
+	store := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
+	bz := store.Get(types.TotalLockupsKey)
+	if bz == nil {
+		return math.ZeroInt()
+	}
+
+	err := total.Unmarshal(bz)
+	if err != nil {
+		return math.ZeroInt()
```
Lodek, line 375: question: do we want to ignore this error? It's unlikely to happen, but if it does it means some code invariant is broken. Would be nice to at least log it

Lodek (approved):
nice! cool optimization

PR #21, opened 2025-02-13
fix: Tier module fixes and improvements (part 6)

iverc:
## Description
This PR adds burn/distribute tier module rewards logic and additional fixes/improvements, following up work done in https://github.com/sourcenetwork/sourcehub/pull/15

## Tasks
- [x] Add custom mint module and custom mint query server.
- [x] Override default `Inflation` query to return effective inflation rate.
- [x] Burn tier module rewards and allocate `2%`/`1%` to `developer_pool`/`insurance_pool`. Rewards that exceed the `InsurancePoolThreshold` are allocated to the `DeveloperPool`. 
- [x] Fix `Lock()` to make sure that `AddLockup()` is called after credits are minted.
- [x] Add `GetTotalLockupsAmount` and `SetTotalLockupsAmount` to keep track of developer stake.
- [x] Add `DeveloperPool` and `InsurancePool` module accounts. 
- [x] Add `DeveloperPoolFee`, `InsurancePoolFee`, and `InsurancePoolThreshold` tier module params.
- [x] Add epochs module basics to fix default `genesis-setup.sh` chain init flow (e.g. epoch info not set).
- [x] Add stub tier module basics to enable overriding default params when needed. 
- [x] Update and fix multiple tests, improve test utils, cleanup.


app/mint.go
```diff
@@ -0,0 +1,179 @@
...
+		return nil, err
+	}
+
+	// Get developer and delegator stake amounts
+	devStake := q.tierKeeper.GetTotalLockupsAmount(ctx)
+	params := q.tierKeeper.GetParams(ctx)
+	totalFees := params.DeveloperPoolFee + params.InsurancePoolFee
+	devStakeMinusFees := devStake.MulRaw(100 - totalFees).QuoRaw(100)
+
+	delStake, err := totalStake.SafeSub(devStakeMinusFees)
```
jsimnz, line 94: suggestion: We should probably have some handy util functions in the tier module to take care of this conversion. So we don't have to keep doing it manually and potentially make mistakes. We can also add some tests to make sure the util is correct, but thats easy.

app/mint.go
```diff
@@ -0,0 +1,179 @@
...
+
+	// If inputs are invalid, return default inflation
+	if !delStake.IsPositive() || !totalStake.IsPositive() {
+		q.logger.Info("Returning default inflation", "inflation", minter.Inflation)
+		return &minttypes.QueryInflationResponse{Inflation: minter.Inflation}, nil
+	}
+
+	// Compute the delegator stake rate and adjusted inflation
+	delStakeRate := delStake.ToLegacyDec().Quo(totalStake.ToLegacyDec())
+	effectiveInflation := minter.Inflation.Mul(delStakeRate)
```
jsimnz: nitpick: Should use `ratio` instead of `rate` terminology.

app/mint.go
```diff
@@ -0,0 +1,179 @@
...
+// RegisterServices registers default message server and custom query server.
+func (cm CustomMintModule) RegisterServices(cfg module.Configurator) {
+	minttypes.RegisterMsgServer(cfg.MsgServer(), mintkeeper.NewMsgServerImpl(cm.mintKeeper))
+	defaultMintQueryServer := mintkeeper.NewQueryServerImpl(cm.mintKeeper)
+	customMintQueryServer := NewCustomMintQueryServer(defaultMintQueryServer, cm.mintKeeper, cm.tierKeeper, cm.tierKeeper.Logger())
+	minttypes.RegisterQueryServer(cfg.QueryServer(), customMintQueryServer)
+}
+
+// InitGenesis initializes default state for the mint keeper.
+func (cm CustomMintModule) InitGenesis(ctx context.Context, cdc codec.JSONCodec, data json.RawMessage) {
```
jsimnz, line 122: question: Is this `InitGenesis` function different from the `InitGenesis` of the original `mint` module? Or, more broadly, what is this custom function doing relative to the default implementation? Lets document it :)

app/mint.go
```diff
@@ -0,0 +1,179 @@
...
+}
+
+// RegisterMintInterfaces registers interfaces for the mint module and returns mint.AppModule.
+func RegisterMintInterfaces(registry codectypes.InterfaceRegistry) appmodule.AppModule {
+	module := mint.AppModule{}
+	if mod, ok := any(module).(interface {
+		RegisterInterfaces(registry codectypes.InterfaceRegistry)
+	}); ok {
+		mod.RegisterInterfaces(registry)
+	}
```
jsimnz, line 176: suggestion: I don't think this interface conversion/casting is necessary. `mint.AppModule` is a concrete struct type and `RegisterInterfaces` is a public method on the type, so you can just call it directly. Any changes will be caught by the compiler.

x/tier/keeper/abci.go
```diff
@@ -0,0 +1,94 @@
...
+	params := k.GetParams(ctx)
+
+	// Process rewards every N blocks
+	height := sdk.UnwrapSDKContext(ctx).BlockHeight()
+	if height%params.ProcessRewardsInterval != 0 {
+		return nil
+	}
+
+	tierModuleAddr := authtypes.NewModuleAddress(types.ModuleName)
+	err := k.GetStakingKeeper().IterateDelegations(ctx, tierModuleAddr, func(index int64, delegation stakingtypes.DelegationI) bool {
```
jsimnz, line 31: suggestion: Although this is obvious if you know how the `IterateDelegations` function or how the staking module stores the delegations works, but lets add a comment here indicating that this iteration loop has a max number of iterations equal to the number of validators. So its bounded :)

x/tier/keeper/abci.go
```diff
@@ -0,0 +1,94 @@
...
+// otherwise the InsurancePoolFee (1%) is also sent to the developer pool.
+// Remaining 97% of the rewards is burned.
+func (k *Keeper) BeginBlocker(ctx context.Context) error {
+	params := k.GetParams(ctx)
+
+	// Process rewards every N blocks
+	height := sdk.UnwrapSDKContext(ctx).BlockHeight()
+	if height%params.ProcessRewardsInterval != 0 {
+		return nil
+	}
```
jsimnz, line 26: praise: Good catch to add this. Hopefully we will be able to measure the perf impact this function has and be able to adjust the interval to accomodate if necessary.

x/tier/keeper/lockup.go
```diff
@@ -241,41 +239,15 @@ func (k Keeper) TotalAmountByAddr(ctx context.Context, delAddr sdk.AccAddress) m
...
-
-	return nil
-}
-
-// IterateUnlockingLockups iterates over all unlocking lockups in the store and performs the provided callback function.
+// iterateUnlockingLockups iterates over all unlocking lockups in the store and performs the provided callback function.
 // The iterator itself doesn't return an error, but the callback does.
 // If the callback returns an error, the iteration stops and the error is returned.
-func (k Keeper) IterateUnlockingLockups(ctx context.Context,
+func (k Keeper) iterateUnlockingLockups(ctx context.Context,
```
jsimnz, line 250: suggestion: Iteration fcuntions are probably fine to keep public :)

x/tier/keeper/lockup.go
```diff
@@ -294,15 +266,15 @@ func (k Keeper) IterateUnlockingLockups(ctx context.Context,
...
 		}
 	}
 
 	return nil
 }
 
-// MustIterateLockups iterates over all lockups in the store and performs the provided callback function.
-func (k Keeper) MustIterateLockups(ctx context.Context,
+// mustIterateLockups iterates over all lockups in the store and performs the provided callback function.
+func (k Keeper) mustIterateLockups(ctx context.Context,
```
jsimnz, line 277: suggestion: Same as above

x/tier/keeper/lockup.go
```diff
@@ -318,8 +290,8 @@ func (k Keeper) MustIterateLockups(ctx context.Context,
 	}
 }
 
-// MustIterateUnlockingLockups iterates over all unlocking lockups in the store and performs the provided callback function.
-func (k Keeper) MustIterateUnlockingLockups(ctx context.Context,
+// mustIterateUnlockingLockups iterates over all unlocking lockups in the store and performs the provided callback function.
+func (k Keeper) mustIterateUnlockingLockups(ctx context.Context,
```
jsimnz, line 294: suggestion: Same as above :)

jsimnz (approved):
Really great stuff! Marking as approved. Theres some `suggestions` and a `question` that could potentially turn into `todos` but at the moment they aren't blockers to the PR.

PR #22, opened 2025-02-19
fix: Panic after updating ibc-go to v8.3.1+

iverc:
## Description

This PR fixes `panic: query router must not be nil` that was introduced after changes made in `ibc-go v8.3.1`.


jsimnz (approved):
ezpz

PR #23, opened 2025-02-23
chore: fix some function names in comment

beforetech:


jsimnz:
@beforetech Thanks for the PR, at the moment we aren't taking outside contributions for small changes like this. If there's something more significant that catches your eye, we are more than happy to accept outside contributions for it.

PR #26, opened 2025-03-02
feature: Block SDK integration

iverc:
## Description

This PR integrates Block SDK, introducing the laned mempool, priority lane, and prepare/process proposal handlers.

Priority lane occupies 100% of the block space and matches all transactions. Transactions within priority lane are prioritized depending on the message types (e.g. `x/acp` > `x/tier` > `x/bulletin` > `other`).


app/ante/ante.go
```diff
@@ -46,6 +49,7 @@ func NewAnteHandler(
 		ante.NewIncrementSequenceDecorator(accountKeeper),
 		// Checks that the tx is not a duplicate IBC packet or update message.
 		ibcante.NewRedundantRelayDecorator(channelKeeper),
+		// auctionante.NewAuctionDecorator(auctionkeeper, TxEncoder, mevLane),
```
Lodek: todo: is this meant to be commented? we can remove otherwise

app/ante/ante.go
```diff
@@ -20,6 +20,8 @@ func NewAnteHandler(
 	TxEncoder sdk.TxEncoder,
 ) sdk.AnteHandler {
 	return sdk.ChainAnteDecorators(
+		// Catches panics and logs errors with the transaction data.
```
Lodek: is this comment accurate? It seems the implemented decorator re-raises the panic.

app/ante/panic.go
```diff
@@ -0,0 +1,32 @@
+package ante
+
+import (
+	"fmt"
+
+	sdk "github.com/cosmos/cosmos-sdk/types"
+)
+
+type HandlePanicDecorator struct{}
```
Lodek, line 10: question: can you comment the purpose of this decorator? The name makes it sound like it catches a panic instead of re-raising it

app/lanes.go
```diff
@@ -0,0 +1,93 @@
...
+func CreateLanes(app *App) (priorityLane *base.BaseLane, defaultLane *base.BaseLane) {
+	// signerAdapter is used to extract the expected signers from a transaction
+	signerAdapter := signerextraction.NewDefaultAdapter()
+
+	// Create a priority lane configuration that occupies 40% of the block space
+	priorityLaneConfig := base.LaneConfig{
+		Logger:          app.Logger(),
+		TxEncoder:       app.txConfig.TxEncoder(),
+		TxDecoder:       app.txConfig.TxDecoder(),
+		MaxBlockSpace:   math.LegacyMustNewDecFromStr("0.4"),
```
Lodek: suggestion: It might be worthwhile to move the constant value to a constants file / params or receive it as an argument to the CreateLanes function. Something like `priorityLaneBlockSpace` and have `defaultLaneBlockSpace = 1 - priorityLaneBlockSpace`.

app/lanes.go
```diff
@@ -0,0 +1,93 @@
...
+	signerextraction "github.com/skip-mev/block-sdk/v2/adapters/signer_extraction_adapter"
+	"github.com/skip-mev/block-sdk/v2/block/base"
+	defaultlane "github.com/skip-mev/block-sdk/v2/lanes/base"
+	"github.com/sourcenetwork/sourcehub/app/lanes"
+)
+
+// CreateLanes creates and returns two lanes: priorityLane and defaultLane.
+// Priority lane uses PriorityMatchHandler to match all transactions related to the x/acp and x/tier modules.
+// Default lane is the default BaseLane implementation.
+func CreateLanes(app *App) (priorityLane *base.BaseLane, defaultLane *base.BaseLane) {
```
Lodek: question: this is not beign used at the moment, right? Do we have plans to use this?

app/lanes/priority_lane.go
```diff
@@ -0,0 +1,148 @@
...
+// getPriorityGroup returns a string that defines the transaction priority.
+// Prioritizes transactions based on the modules their messages belong to.
+func getPriorityGroup(tx sdk.Tx) string {
+	msgs := tx.GetMsgs()
+	if len(msgs) == 0 {
+		return "0"
+	}
+	// Start with highest possible priority and reduce txPriority based on the found messages,
+	// so that the system can not be abused by sending a mix or high and low priority messages
+	minPriority := "3"
```
Lodek, line 74: discussion: This is interesting, I think I see why you did it, there are two scenarios though.

case 1: Suppose someone has a bulletin msg that they want to be processed before some other msg, by appending *any* ACP msg (invalid or otherwise) to their Tx they can get priority, which is an exploit.

case 2: suppose a scenario where a Tx has a valid ACP Tx and some other messages, the Tx should get priority because of the ACP message.

I don't think there's a right answer here, only a choice on what we want to optimize for, however I do think that this behavior should be explicitly documented somewhere.

Lodek (approved):
Good job! I don't want to block the PR if the points I brought up have already been considered but there are a couple of pointers which I think are worth noting, such as the `minPriority` system

app/lanes/priority_lane.go
iverc: @Lodek the primary reason for this implementation is to prevent the exploit described in case 1: 
- If a transaction includes a Bulletin message, adding an ACP or Tier message will not boost its priority. The logic ensures that the transaction’s priority is always determined by the lowest priority message.

Regarding case 2:
- If the goal is to make sure that your ACP message is prioritized correctly, the best approach is to send it as a standalone transaction rather than bundling it with other messages. In most real-world cases, a well-intentioned client would already be structuring txs this way.

app/lanes.go
iverc: yes, we can remove it

app/ante/ante.go
iverc: thanks, fixed the comment

app/ante/ante.go
iverc: good catch, removed

app/lanes/priority_lane.go
Lodek: Sounds reasonable. Also it's worth noting that your approach is more inclusive. By having the code as it is, it's possible to catch bad-actors, while it would not be possible otherwise. 
I'm happy to go along with this, just think we should document it under source's main docs at some point as it could be easily overlooked.

iverc: Makes sense, thank you.

PR #27, opened 2025-03-03
refactor: Delegator stake ratio calculation  

iverc:
## Description

Refactor the delegator stake ratio calculation into a standalone function and add corresponding test cases.

### DIFF
--- a/app/mint.go
+++ b/app/mint.go
@@ -3,6 +3,7 @@ package app
 import (
 	"context"
 	"encoding/json"
+	"fmt"
 
 	"cosmossdk.io/log"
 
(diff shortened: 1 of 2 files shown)


app/mint.go
```diff
@@ -71,39 +72,51 @@ func NewCustomMintQueryServer(
...
+
+	if !totalStake.IsPositive() || !delStake.IsPositive() {
+		return math.LegacyOneDec(), fmt.Errorf("non-positive totalStake/delStake")
+	}
+
+	// Calculate the delegator stake ratio
+	delStakeRatio := delStake.ToLegacyDec().Quo(totalStake.ToLegacyDec())
+
+	return delStakeRatio, nil
+}
```
Lodek, line 102: suggestion: keep this function pure, and receive the values fetched from the keeper as args. Maybe that would allow merging the two funcs again.

eg.
```
func getDelegatorStakeRatio(developerPoolFee, insurancePoolFee, devStake math.LegacyDev) (math.LegacyDec, error) {
```

Lodek (approved):
lgmt, just one refactor suggestion

PR #28, opened 2025-03-04
feature: Enable ucredit fees

iverc:
## Description

This PR enables fees in `ucredit` denom. 

The minimum gas price setting is identical for both `uopen` and `ucredit` denoms.

### DIFF
--- a/scripts/genesis-setup.sh
+++ b/scripts/genesis-setup.sh
@@ -1,7 +1,7 @@
 #!/usr/bin/sh
 set -e
 
-CHAIN_ID="sourcehub"
+CHAIN_ID="sourcehub-dev"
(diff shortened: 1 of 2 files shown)


scripts/genesis-setup.sh
```diff
@@ -1,7 +1,7 @@
 #!/usr/bin/sh
 set -e
 
-CHAIN_ID="sourcehub-dev"
```
Lodek: can we keep the `-dev` suffix? I don't know what name our mainnet will take and I'd rather avoid the chance of a name clash for whatever reason

Lodek (approved):
lgtm, just one nitpick

PR #31, opened 2025-03-07
feat: Tier module slashing insurance

iverc:
## Description

This PR adds changes to monitor `slash` events on every block and cover the `x/tier` module share of the burned/slashed tokens as long as the reason was `missing_signature`, as well as changes to handle `double_sign` events to update existing records to reflect actual amounts after slashing and additional fixes/refactoring.


x/tier/keeper/abci_test.go
```diff
@@ -42,3 +47,77 @@ func TestBeginBlocker(t *testing.T) {
...
+
+	err = k.handleSlashingEvents(ctx)
+	require.NoError(t, err)
+
+	newBalance = k.GetBankKeeper().GetBalance(ctx, insurancePoolAddr, appparams.DefaultBondDenom)
+	require.Equal(t, expectedRemainingBalance, newBalance.Amount)
+
+	tierDelegationAfter, err := k.GetStakingKeeper().GetDelegation(ctx, tierModuleAddr, valAddr)
+	require.NoError(t, err)
+	require.Equal(t, tierDelegation.Shares, tierDelegationAfter.Shares)
```
Lodek: question: would it make sense to also test the delegated amount of the validator after the recovery? It's my understanding that the tier module delegates the recovered amount back to the punished validator, right?

x/tier/keeper/abci.go
```diff
@@ -2,21 +2,126 @@ package keeper
...
+	}
+
+	// Get the total stake of the slashed validator
+	totalStake := validator.Tokens.ToLegacyDec()
+	if totalStake.IsZero() {
+		return fmt.Errorf("No stake for the validator: %s", validatorAddr)
+	}
+
+	// Get burned (slashed) amount
+	totalBurned, err := math.LegacyNewDecFromStr(burnedAmount)
```
Lodek: question: I would like to understand this bit of the code better, as I am not too familiar with the token economics stuff. What's the difference between recovering the tierShare and the totalBurned? is it because we are only recovering an amount proportial to a lockup or something?

Lodek (approved):
Looks good to me, just one comment about a test and a question to understand the tier module better. Good job!

x/tier/keeper/abci.go
iverc: @Lodek right, we are only recovering the tier module portion of the slashed stake, normal delegator stake is not reimbursed

x/tier/keeper/abci.go
```diff
@@ -2,94 +2,18 @@ package keeper
...
 func (k *Keeper) BeginBlocker(ctx context.Context) error {
-	params := k.GetParams(ctx)
-
-	// Process rewards every N blocks
-	height := sdk.UnwrapSDKContext(ctx).BlockHeight()
-	if height%params.ProcessRewardsInterval != 0 {
-		return nil
+	err := k.handleSlashingEvents(ctx)
+	if err != nil {
+		k.Logger().Error("Failed to handle slashing event", "error", err)
```
Lodek: we should emit a special metric for these internal errors which get ignored (eg sourcehub_errors_internal_total) . Can be done after.

x/tier/keeper/abci.go
```diff
@@ -2,94 +2,18 @@ package keeper
...
-		}
-
-		return false
-	})
-
+	err = k.processRewards(ctx)
 	if err != nil {
-		k.Logger().Error("Error iterating over tier module delegations", "error", err)
-		return err
+		k.Logger().Error("Failed to process rewards", "error", err)
```
Lodek, line 13: same here

x/tier/keeper/slashing.go
```diff
@@ -0,0 +1,187 @@
...
+	appparams "github.com/sourcenetwork/sourcehub/app/params"
+	"github.com/sourcenetwork/sourcehub/x/tier/types"
+)
+
+// handleSlashingEvents monitors and handles slashing events.
+// In case of double_sign, existing lockup records are updated to reflect changes after slashing.
+// Otherwise, in addition to updating existing lockup records, slashed tokens are covered via insurance lockups.
+func (k *Keeper) handleSlashingEvents(ctx context.Context) error {
+	sdkCtx := sdk.UnwrapSDKContext(ctx)
+	events := sdkCtx.EventManager().Events()
```
Lodek, line 22: suggestion: comment somewhere in the tier module that its being blocker must be executed after the slashing module's being blocker because it consumes events processed by the slashing module

x/tier/keeper/slashing.go
```diff
@@ -0,0 +1,187 @@
...
+					slashedAmount = string(attr.Value)
+				}
+			}
+
+			if reason == slashingtypes.AttributeValueDoubleSign {
+				return k.handleDoubleSign(ctx, validatorAddr, slashedAmount)
+			} else {
+				return k.handleMissingSignature(ctx, validatorAddr, slashedAmount)
+			}
+		}
```
Lodek: question: is it possible for multiple slashing events to be emitted for the same begin blocker? Right now we only handle the first one, then return. Is that the desired behavior?

x/tier/keeper/rewards.go
```diff
@@ -0,0 +1,91 @@
...
+
+	tierModuleAddr := authtypes.NewModuleAddress(types.ModuleName)
+	// Iterate over all active delegations where the tier module account is the delegator
+	// The max number of iterations is the number of validators it has delegated to
+	err := k.GetStakingKeeper().IterateDelegations(ctx, tierModuleAddr, func(index int64, delegation stakingtypes.DelegationI) bool {
+		// Claim rewards for the tier module from this validator
+		valAddr := types.MustValAddressFromBech32(delegation.GetValidatorAddr())
+		rewards, err := k.GetDistributionKeeper().WithdrawDelegationRewards(ctx, tierModuleAddr, valAddr)
+		if err != nil {
+			k.Logger().Error("Failed to claim tier module staking rewards", "error", err)
```
Lodek, line 36: note: increment internal error counter, once we add that

Lodek (commented):
Good job Ivan.

My primary comment is regarding multiple stalshing events per block, everything else looks good.

x/tier/keeper/slashing.go
iverc: Nice catch, we should not return there. Thanks!

PR #35, opened 2025-03-17
fix: General improvements

iverc:
## Description

This PR adds changes to improve tier/acp messages and queries, update timeout configs, and cleanup the code.


Lodek (commented):
question: why are we changing back from the querier changes?

Lodek (approved):
:saluting_face:

PR #38, opened 2025-03-25
fix: Acp EndBlock error

iverc:
## Description

This PR refactors acp `EndBlocker` to return `nil` commitments in case of error instead of propagating the error to `EndBlock`.


jsimnz (approved):
LGTM

PR #40, opened 2025-04-02
feat: Telemetry

iverc:
## Description

This PR enables telemetry and prometheus in the config, and adds instrumentation to abci and keeper methods.


Lodek:
I took a look at the PR and I think the cosmos-sdk telemtry package has some design flaws.
Essentially it doesn't allows us to set labels for the metrics and I think that will cause a metric explosion and I just can't get behind it.

The sdk telemetry package uses the `keys` as "hiearchical" / dot-separated names in the underlying hashicorp package lib, with no real option for us to label things accordingly, which IMO hinders how we structure metrics a lot.

One example is that we may want to measure traffic per module. The canonical way would be having a metric called `sourcehub_msg_count`, with labels `msg_name` and `module`.
The cosmos lib doesn't allow that.
Instead we end up creating a metric per msg type which leaves a lot to be desired in terms of query-ability (see final console output of https://pkg.go.dev/github.com/hashicorp/go-metrics#section-readme).

Another example why labels are beneficial is, we could have `sourcehub_msg_execution_seconds` with the labels {module, msg}.
This would allows us to slice the latency per module and message, as we would like.
Having all that information in a *single* metric is beneficial since the queries would be easier. Ex: we can execute an avg / mean for the entire application, for a module or for a message type with a single metric.
The alternative is to "join" all the separated metrics, which would be a pain.

I propose we either use underlying hashicorp lib directly or emit sourcehub specific metrics using open telemetry instead. 
I'm happy to brain storm a few solutions to make things easier as well. 
Perhaps we can inject a metric collector in the suppliers of dep inject, which is already bound to the module name.

I am a fan of keeping things simple, but in this case I think the cosmos pkg would make the metrics simply too messy / verbose to be useful.

app/app.go
```diff
@@ -367,8 +370,29 @@ func New(
...
+	app.App.SetPrepareProposal(wrappedPrepareProposal)
+
+	processProposal := proposalHandler.ProcessProposalHandler()
+	wrappedProcessProposal := func(ctx sdk.Context, req *abcitypes.RequestProcessProposal) (*abcitypes.ResponseProcessProposal, error) {
+		defer telemetry.MeasureSince(time.Now(), metrics.ProcessProposal)
+		telemetry.IncrCounterWithLabels(
+			[]string{metrics.App, metrics.Tx, metrics.Count},
+			float32(len(req.Txs)),
+			[]metrics.Label{telemetry.NewLabel(metrics.Method, metrics.ProcessProposal)},
+		)
```
Lodek: I am a bit confused about this block.
Why are we incrementing the Tx count twice? I understand there is a label for the method, but the global tx counter will still be incremented.

If we indeed want to track tx count for process and prepapre proposal, I think we should have different metrics for it, and leave the main metric `sourcehub_tx_count` metric to be the total number processed by the accepted block because right now, it is twice that.

app/metrics/utils.go
```diff
@@ -0,0 +1,123 @@
...
+		value,
+		labels,
+	)
+}
+
+// ModuleMeasureWithCounter emits latency and success/error counter metrics for a module message with optional labels.
+func ModuleMeasureWithCounter(module, msgType string, start time.Time, err error, extraLabels []Label) {
+	labels := []Label{
+		telemetry.NewLabel(Msg, msgType),
+	}
```
Lodek:
```suggestion
func ModuleMeasureWithCounter(module, msgType string, start time.Time, err error, extraLabels []Label) {
	if !telemetry.IsTelemetryEnabled() {
		return
	}
	labels := []Label{
		telemetry.NewLabel(Msg, msgType),
	}
```

Lodek (approved):

x/tier/module/module.go
```diff
@@ -116,8 +117,11 @@ func NewAppModule(
 
 // RegisterServices registers a gRPC query service to respond to the module-specific gRPC queries
 func (am AppModule) RegisterServices(cfg module.Configurator) {
-	types.RegisterMsgServer(cfg.MsgServer(), keeper.NewMsgServerImpl(&am.keeper))
-	types.RegisterQueryServer(cfg.QueryServer(), &am.keeper)
+	types.RegisterMsgServer(cfg.MsgServer(), keeper.NewMsgServerImpl(am.keeper))
```
Lodek: please still wrap the tier module's service descriptor with the metric injector. Those metrics can be seen as global to sourcehub, we can use them to track the overall latency of all msgs in the system

The descriptor wrappers act as module agnostic metrics that we can use to track the whole application, for tier module specific apps you manually craft metrics with the labels you consider necessary

iverc: @Lodek I believe that would result in double counting same metrics since they are already tracked in each method separately

Lodek: Yes, you're right. I noticed that `ModuleMeasurSinceWithCounter`, adds extra labels to the main app metrics (eg. `SourcehubMsgSeconds`). I think that can be problematic, since those metrics are used by all methods in all modules. Adding extra labels to it creates a new dimension for each label. It also means a record of `SourceHubMsgSeconds` for `module=acp` would have a label `amount=""`. I think these top level metrics, should be general purpose for the primary stats.

My suggestion would be, if you want to track those things independently, I think a separate metric would make more sense.
There's a fine line between sufficiently granular metrics and treating them as logs, but I trust your judgment in that you can see value in tracking that data.

iverc: Yeah, you're right. Thanks for the suggestion. It would make more sense to track that separately as `SourceHubMethodSeconds` and keep the base metric enabled via the wrapper.

app/metrics/utils.go
```diff
@@ -0,0 +1,119 @@
...
+
+type Label = gometrics.Label
+
+// NewLabel creates new Label with given name and value.
+func NewLabel(name, value string) Label {
+	return Label{Name: name, Value: value}
+}
+
+// ModuleMeasureSinceWithLabels emits a time measure metric for a module with a given set of keys and labels.
+func ModuleMeasureSinceWithLabels(module string, start time.Time, keys []string, extraLabels []Label) {
```
jsimnz: question: Why do these first two functions `ModuleMeasureSinceWithLabels` and `ModuleIncrCounterWithLabels` not start with the same telemetry enabled check?

x/tier/keeper/keeper.go
```diff
@@ -146,7 +159,23 @@ func (k *Keeper) CompleteUnlocking(ctx context.Context) error {
...
 
 // Lock locks the stake of a delegator to a validator.
-func (k *Keeper) Lock(ctx context.Context, delAddr sdk.AccAddress, valAddr sdk.ValAddress, amt math.Int) error {
+func (k *Keeper) Lock(ctx context.Context, delAddr sdk.AccAddress, valAddr sdk.ValAddress, amt math.Int) (err error) {
+	start := time.Now()
+
+	defer func() {
+		metrics.ModuleMeasureSinceWithCounter(
+			types.ModuleName,
+			metrics.Lock,
```
jsimnz, line 168: suggestion: This is somewhat general to this PR, but highlighted here. Imo, the actual metrics instance (ie: `metrics.Lock`) should be defined within the respective modules. Otherwise, `metrics.Lock` isn't very descriptive without further context (A fuller name would be something like `metrics.TierDevLock`). If they were defined within the modules, for examplein the module `types` folder, it might make them a little clearer without need to be overly specific/descriptive for each name.

Not a blocker for this PR though.

jsimnz (approved):
LGTM. Love the specific instrumentation on the `tier` module.

Minor suggestion naming/location of the metric types, but approving now regardless.

app/metrics/utils.go
iverc: I moved that check to `ModuleMeasureMsgWithCounter` in the latest commit since `ModuleMeasureSinceWithLabels` and `ModuleIncrCounterWithLabels` are only called within it. However, I'm actually going to remove all three methods because none of them is used atm.

PR #42, opened 2025-04-02
refactor: Encapsulate ACP Keeper methods

Lodek:
What:
This PR refactors several methods in the ACP Keeper to private methods.

Why:
The original design of the keeper had an oversight where the usage of the Keeper in different SourceHub modules was not considered.
As such, this left several internal components of the ACP Module public to external modules, which broke encapsulation and incorrect usage of the ACP methods could lead to security compromises of the Policies in ACP


iverc (approved):
LGTM 👍

PR #45, opened 2025-04-23
feat: Scripts

iverc:
## Description

This PR adds utility scripts for local chain setups and updates/improves existing scripts.

### DIFF
--- a/scripts/genesis-setup.sh
+++ b/scripts/genesis-setup.sh
@@ -1,26 +1,42 @@
-#!/usr/bin/sh
+#!/bin/sh
 set -e
+
+rm -rf ~/.sourcehub || true
+
+sedi() {
+  if [[ "$OSTYPE" == "darwin"* ]]; then
+    sed -i '' "$@"
+  else
+    sed -i "$@"
+  fi
+}
 
 CHAIN_ID="sourcehub-dev"
 VALIDATOR="validator"
 NODE_NAME="node"
 BIN="build/sourcehubd"
 
-rm -rf ~/.sourcehub || true
-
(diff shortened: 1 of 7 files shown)


scripts/genesis-setup.sh
```diff
@@ -1,26 +1,42 @@
...
+sedi 's/minimum-gas-prices = .*/minimum-gas-prices = "0.001uopen,0.001ucredit"/' ~/.sourcehub/config/app.toml
+sedi 's/^enabled = .*/enabled = true/' ~/.sourcehub/config/app.toml
+sedi 's/^prometheus-retention-time = .*/prometheus-retention-time = 60/' ~/.sourcehub/config/app.toml
+
+# config.toml
+sedi 's/^timeout_propose = .*/timeout_propose = "500ms"/' ~/.sourcehub/config/config.toml
+sedi 's/^timeout_prevote = .*/timeout_prevote = "500ms"/' ~/.sourcehub/config/config.toml
+sedi 's/^timeout_precommit = .*/timeout_precommit = "500ms"/' ~/.sourcehub/config/config.toml
+sedi 's/^timeout_commit = .*/timeout_commit = "1s"/' ~/.sourcehub/config/config.toml
+sedi 's/^prometheus = .*/prometheus = true/' ~/.sourcehub/config/config.toml
```
Lodek, line 40: I wonder whether we can change the defaults of some of these options on the SourceHub binary, so that we don't need to change these same configs in a bunch of scripts.

Lodek (approved):

scripts/genesis-setup.sh
iverc: The same binary could be used for different deployments so I think it makes more sense to set these options in the configs

PR #46, opened 2025-04-25
feat: Zero Fees

iverc:
## Description

This PR adds zero-fee transactions param to genesis. If zero-fee transactions are allowed and `--fees` flag is omitted, the ante handler skips fee verification logic. Otherwise, sufficient fee is required for transaction to succeed.

### DIFF
--- a/app/app.go
+++ b/app/app.go
@@ -65,6 +65,7 @@ import (
 	"github.com/sourcenetwork/sourcehub/app/ante"
 	"github.com/sourcenetwork/sourcehub/app/metrics"
 	overrides "github.com/sourcenetwork/sourcehub/app/overrides"
+	appparams "github.com/sourcenetwork/sourcehub/app/params"
 	sourcehubtypes "github.com/sourcenetwork/sourcehub/types"
 	acpkeeper "github.com/sourcenetwork/sourcehub/x/acp/keeper"
 	acptypes "github.com/sourcenetwork/sourcehub/x/acp/types"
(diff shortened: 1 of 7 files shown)


app/app.go
```diff
@@ -432,6 +434,21 @@ func New(
...
+			if err := json.Unmarshal(raw, &appParams); err == nil {
+				store := ctx.KVStore(app.GetKey(authtypes.StoreKey))
+				val := byte(0x00)
+				if appParams.AllowZeroFeeTxs {
+					val = 0x01
+				}
+				store.Set([]byte(appparams.AllowZeroFeeTxsKey), []byte{val})
+			}
+		}
+
```
Lodek: suggestion: IMO we should set the value in the store even if the genesis param is not defined.
```suggestion
		// Parse app_state.app_params.allow_zero_fee_txs and save to auth (acc) store.
		val := byte(0x00)
		if raw, ok := genesisState["app_params"]; ok {
			var appParams struct {
				AllowZeroFeeTxs bool `json:"allow_zero_fee_txs"`
			}
			if err := json.Unmarshal(raw, &appParams); err == nil  && appParams.AllowZeroFeeTxs{
				val = 0x01
			}
		}
		store := ctx.KVStore(app.GetKey(authtypes.StoreKey))
		store.Set([]byte(appparams.AllowZeroFeeTxsKey), []byte{val})



```

app/app.go
```diff
@@ -432,6 +434,21 @@ func New(
 			return nil, err
 		}
 
+		// Parse app_state.app_params.allow_zero_fee_txs and save to auth (acc) store.
+		if raw, ok := genesisState["app_params"]; ok {
+			var appParams struct {
```
Lodek: suggestion: I think we can define this struct outside the init chainer. As of right now, this behavior is pretty obscure. Having this struct defined in a constants file in app would more clearly document this feature.

Lodek (approved):
Good job. Couple of suggestions, non blocking.

PR #47, opened 2025-04-28
fix: Move telemetry check for wrappers

Lodek:
Fixes issue where metrics weren't emitted for module wrappers.

### DIFF
--- a/app/metrics/wrapper.go
+++ b/app/metrics/wrapper.go
@@ -50,23 +50,25 @@ func wrapMsgSeverHandler(
 		dec func(interface{}) error,
 		interceptor grpc.UnaryServerInterceptor,
 	) (interface{}, error) {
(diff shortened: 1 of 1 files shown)


app/metrics/wrapper.go
```diff
@@ -64,15 +56,17 @@ func wrapMsgSeverHandler(
 		}
 		labels = append(labels, commonLabels...)
 
-		// Track message handling latency
-		now := time.Now()
-		defer gometrics.MeasureSinceWithLabels(latencyMetricName, now, labels)
+		if telemetry.IsTelemetryEnabled() {
```
iverc: question: can we have the `!telemetry.IsTelemetryEnabled()` check in the beginning of the function?

iverc (approved):

PR #50, opened 2025-05-05
feat: Testnet cmd

iverc:
## Description

This PR adds `testnet` cli command with two subcommands: `init-files` and `start`. 

- `testnet init-files` initializes config directories and files for a multi-validator testnet setup, intended for use with Docker Compose.
- `testnet start` launches an in-process multi-validator testnet directly from the CLI.


Lodek (approved):
wow, really cool. Good job Ivan

PR #51, opened 2025-05-06
fix: Tx Listener cleanup and fix

Lodek:
What:

This PR introduces fixes a bug in the Tx Listener, adds tests and cleans up the listener cmd


Dockerfile
```diff
@@ -1,25 +1,4 @@
-FROM golang:1.23.8 as builder
```
iverc: I think we might wanna have a separate dockerfile for the tx listener

Lodek: I think you're right. After double checking the binary size, it's surprisingly large (140M). I'll revise that

Lodek: Will move this utility to a new repo

iverc (approved):
LGTM

PR #54, opened 2025-05-29
ci: Cleanup dockerfile and add /sourcehub as data volume

Lodek:


jsimnz:
@Lodek whats the TLDR for the need for this PR?

Lodek:
> @Lodek whats the TLDR for the need for this PR?

- Removes some extra steps in the build from back when I first wrote which used ignite to build (reduce bloat and speed up)
- Primary change is the volume directive which is nicer to run locally since it auto creates a volume
- Adds /sourcehub as the data directory, which is linked to ~/.sourcehub
- Overall just makes it nicer to deal with the node state while writing composes and manifests

jsimnz (approved):
Lgtm

PR #61, opened 2025-07-22
feat: Dev faucet

iverc:
## Description

This PR adds built-in dev faucet with multiple endpoints, fixes existing faucet script, and improves genesis-setup script.

Supported faucet endpoints:

- `/faucet/info` - get faucet info like balance and request count.
- `/faucet/request` - request funds from the faucet, limited to 1000 `$OPEN` per address.
- `/faucet/init-account` - initialize new account in the `auth` module by sending 1 `uopen`.

Additional fixes:
- Fix `./scripts/faucet.sh` to use `uopen` and allow requesting arbitrary token amount from the faucet. 
- Improve `./scripts/genesis-setup.sh` to enable API, allow CORS, and have faucet with zero-fees enabled by default.


app/faucet.go
```diff
@@ -0,0 +1,426 @@
...
+		}
+
+		_, err := sdk.AccAddressFromBech32(req.Address)
+		if err != nil {
+			http.Error(w, "Invalid address", http.StatusBadRequest)
+			return
+		}
+
+		// Check if address has already requested funds
+		if app.hasAddressRequested(req.Address) {
```
Lodek, line 187: is this check necessary? what issues could arise from letting the same acc request more than once?

app/faucet.go
```diff
@@ -0,0 +1,426 @@
...
+			return
+		}
+
+		amount := "1000000000uopen" // 1000 open
+
+		coins, err := sdk.ParseCoinsNormalized(amount)
+		if err != nil {
+			http.Error(w, "Invalid amount", http.StatusBadRequest)
+			return
+		}
```
Lodek: suggestion: non-blocking but since this is static we could do this in the init and get rid of this block altogether

app/faucet.go
```diff
@@ -0,0 +1,426 @@
...
+		msg := banktypes.NewMsgSend(
+			faucetAddress,
+			sdk.MustAccAddressFromBech32(req.Address),
+			coins,
+		)
+
+		chainID := clientCtx.ChainID
+		if chainID == "" {
+			chainID = "sourcehub-dev"
+		}
```
Lodek, line 215: suggestion: would it make sense to use the client from the `sdk` package to reduce some of the tx signing / broadcasting boilerplate?

app/faucet.go
```diff
@@ -0,0 +1,426 @@
...
+
+		// Create a bank send message with 1 uopen to initialize the account
+		coins := sdk.NewCoins(sdk.NewInt64Coin("uopen", 1))
+		msg := banktypes.NewMsgSend(
+			faucetAddress,
+			accAddr,
+			coins,
+		)
+
+		chainID := clientCtx.ChainID
```
Lodek, line 370: question: same observation about the sdk pkg as before

Lodek (commented):
Overall it looks good, thanks for the work Ivan.

I do have a slight preference about defining the API request / responses msgs as protobuff messages, instead of the map approach. I think doing so would clean up both the faucet code but also the tests.
Furthermore, through the proto definition we might be able to leverage the open api code generation and add it to the swagger interface for the api.
I am happy to have a chat about that if you feel differently.

app/faucet.go
iverc: this check is not strictly necessary for local deployments, but if it is used for devnet, without it a single address can drain the faucet and spam the network with lots of txs.

Lodek: alright, if we see a need for it we can remove it in the future, since in the devnet we probably want some sort of rate-limiting proxy in front of the rpcs

Lodek (approved):

PR #63, opened 2025-08-01
fix: Create policy cli command

iverc:
## Description

This PR fixes the number of arguments in the `create-policy` cli command to allow setting the `marshaling_type`.


Lodek (approved):
lgtm

PR #66, opened 2025-08-14
feat: IBC ACP (v1)

iverc:
## Description

This PR enables ICA host, allows executing ACP messages from a controller chain via ICA, and adds basic `IssueInterchainAccountDID` implementation. 

Additionally, adds a helper to generate ICA packets and a script to demo ACP policy creation with ICA.

### DIFF
--- a/app/ibc.go
+++ b/app/ibc.go
@@ -174,7 +174,18 @@ func (app *App) setDefaultIBCParams(ctx sdk.Context) {
 	app.IBCKeeper.ClientKeeper.SetParams(ctx, ibcclienttypes.DefaultParams())
 	app.IBCKeeper.ConnectionKeeper.SetParams(ctx, ibcconnectiontypes.DefaultParams())
 	app.ICAControllerKeeper.SetParams(ctx, icacontrollertypes.DefaultParams())
-	app.ICAHostKeeper.SetParams(ctx, icahosttypes.DefaultParams())
+	// Enable ICA host and allow ACP msgs to be executed from interchain accounts
+	hostParams := icahosttypes.DefaultParams()
+	hostParams.HostEnabled = true
+	hostParams.AllowMessages = []string{
+		"/sourcehub.acp.MsgCreatePolicy",
+		"/sourcehub.acp.MsgEditPolicy",
+		"/sourcehub.acp.MsgCheckAccess",
+		"/sourcehub.acp.MsgSignedPolicyCmd",
+		"/sourcehub.acp.MsgBearerPolicyCmd",
+		"/sourcehub.acp.MsgDirectPolicyCmd",
+	}
+	app.ICAHostKeeper.SetParams(ctx, hostParams)
 	app.TransferKeeper.SetParams(ctx, ibctransfertypes.DefaultParams())
 }
 
--- a/scripts/start-multichain-ica.sh
+++ b/scripts/start-multichain-ica.sh
@@ -62,7 +62,7 @@ mkdir $HERMES_HOME
 cp scripts/hermes_config.toml "$HERMES_HOME/config.toml"
 
 # Build the binary
-make build-mac # make build-mac
+make build # make build-mac
 
 echo "==> Initializing sourcehub-1..."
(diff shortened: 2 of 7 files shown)


app/ibc.go
```diff
@@ -174,7 +174,18 @@ func (app *App) setDefaultIBCParams(ctx sdk.Context) {
...
+	hostParams := icahosttypes.DefaultParams()
+	hostParams.HostEnabled = true
+	hostParams.AllowMessages = []string{
+		"/sourcehub.acp.MsgCreatePolicy",
+		"/sourcehub.acp.MsgEditPolicy",
+		"/sourcehub.acp.MsgCheckAccess",
+		"/sourcehub.acp.MsgSignedPolicyCmd",
+		"/sourcehub.acp.MsgBearerPolicyCmd",
+		"/sourcehub.acp.MsgDirectPolicyCmd",
+	}
```
reecepbcups, line 187: no action required
Just want to call out that `[]string{"*"}` is also a valid syntax if you want to support all messages on the network (and is used in some production networks)

x/acp/keeper/utils.go
```diff
@@ -23,6 +23,11 @@ func (k *Keeper) IssueDIDFromAccountAddr(ctx context.Context, addr string) (stri
 		return "", fmt.Errorf("IssueDIDFromAccountAddr: %w", types.NewAccNotFoundErr(addr))
 	}
 
+	// Interchain Accounts (ICA) created by the host module have no pubkey set.
+	if acc.GetPubKey() == nil {
+		return did.IssueInterchainAccountDID(sdkAddr.String()), nil
+	}
```
Lodek, line 29: I think this approach can work for the time being, but do you see another approach which doesn't rely on checking if the pubkey is nil?

I am thinking primarily from a future proofing perspective. Currently dids are only issued to Tx signers, which automatically means they have the pub key populated by the antehandler. However, if at some point the acp module starts issueing dids to some other target account which doesn't sign the tx, the account might have been created through the bank module, but never signed a Tx, hence pub key is nil.

scripts/start-multichain-ica.sh
```diff
@@ -0,0 +1,219 @@
...
+$BIN genesis validate-genesis --home "$C1V1_HOME"
+
+jq '.app_state.transfer.port_id = "transfer"' "$C1V1_HOME/config/genesis.json" > tmp.json && mv tmp.json "$C1V1_HOME/config/genesis.json"
+jq '.app_state.transfer += {"params": {"send_enabled": true, "receive_enabled": true}}' "$C1V1_HOME/config/genesis.json" > tmp.json && mv tmp.json "$C1V1_HOME/config/genesis.json"
+
+sed -i '' 's/minimum-gas-prices = ""/minimum-gas-prices = "0.001uopen,0.001ucredit"/' "$C1V1_HOME/config/app.toml"
+sed -i '' 's/^timeout_propose = .*/timeout_propose = "500ms"/' "$C1V1_HOME/config/config.toml"
+sed -i '' 's/^timeout_prevote = .*/timeout_prevote = "500ms"/' "$C1V1_HOME/config/config.toml"
+sed -i '' 's/^timeout_precommit = .*/timeout_precommit = "500ms"/' "$C1V1_HOME/config/config.toml"
+sed -i '' 's/^timeout_commit = .*/timeout_commit = "1s"/' "$C1V1_HOME/config/config.toml"
```
Lodek, line 86: question: (no action requried) we are starting to have a number of scripts to spin up different environment combinations with a fair amount of repetition. I've been thinking about ways to improve that. the operator work could come into play. in the future, would you be opposed to running local soucehub deployments like these through docker instead?

scripts/start-multichain-ica.sh
```diff
@@ -0,0 +1,219 @@
+#!/bin/sh
```
Lodek, line 1: suggestion: we can migrate the previous binary and this script into  `tests/e2e/ica`. That'd make a bit easier to understand how things are tested

x/acp/keeper/utils.go
iverc: yes, we can definitely have more robust approach for v2.

Lodek (approved):

PR #67, opened 2025-08-14
feat: Return parsed Policy in ValidatePolicy query

Lodek:
What:
Add an extra field to the response payload of the `ValidatePolicy` query, which returns the parsed Policy.
Employs some minor refactors from the latest acp_core bump

Why:
The DB requested a way to programatically inspect a Policy, before it gets created.
The Validate method went through all the parse and validation logic as it is, therefore it made sense to simply return the parsed policy


x/acp/keeper/query_registrations_commitment.go
```diff
@@ -27,7 +28,8 @@ func (k *Keeper) RegistrationsCommitment(
 	}
 	if opt.Empty() {
 		return nil, errors.Wrap("commitment not found", errors.ErrorType_NOT_FOUND,
-			errors.Pair("commitment", req.Id))
+			errors.Pair("commitment", fmt.Sprintf("%v", req.Id)),
```
iverc, line 31: Is this change required?

x/acp/types/errors.go
```diff
@@ -25,11 +25,11 @@ const (
 
 var New = errors.New
 var Wrap = errors.Wrap
-var NewFromBaseError = errors.NewFromBaseError
-var NewFromCause = errors.NewFromCause
+var NewFromBaseError = errors.NewWithCause
```
iverc: It looks like both `NewFromCause ` and `NewFromBaseError` are set to `NewWithCause`, should we keep `var NewFromBaseError = errors. NewFromBaseError`?

x/acp/registration/service.go
```diff
@@ -179,14 +181,14 @@ func (s *RegistrationService) FlagHijackEvent(ctx sdk.Context, eventId uint64, a
...
-		return nil, errors.Wrap("event not found", errors.ErrorType_NOT_FOUND, errors.Pair("event", eventId))
+		return nil, errors.Wrap("event not found", errors.ErrorType_NOT_FOUND, errors.Pair("event", fmt.Sprintf("%v", eventId)))
 	}
 	event := opt.GetValue()
 
 	if event.Metadata.OwnerDid != actor.Id {
 		return nil, errors.Wrap("event actor missmatch: actor must be event subject",
 			errors.ErrorType_UNAUTHORIZED,
-			errors.Pair("event", eventId),
+			errors.Pair("event", fmt.Sprintf("%v", eventId)),
```
iverc, line 191: same here

x/acp/commitment/service.go
```diff
@@ -91,7 +93,7 @@ func (s *CommitmentService) FlagExpiredCommitments(ctx sdk.Context) ([]*types.Re
 	for _, commitment := range processed {
 		err := s.repository.update(ctx, commitment)
 		if err != nil {
-			return nil, errors.Wrap("expiring commitment", err, errors.Pair("commitment", commitment.Id))
+			return nil, errors.Wrap("expiring commitment", err, errors.Pair("commitment", fmt.Sprintf("%v", commitment.Id)))
```
iverc, line 96: same here

iverc (approved):
A couple of questions, other than that LGTM.

x/acp/keeper/query_registrations_commitment.go
Lodek: yes, acp_core expects a string, string pair. Honestly, I'll prob revert in the future to accept any but outside of the scope of this PR for now

x/acp/types/errors.go
Lodek: Good catch, I meant to remove one of them because they were redundant

PR #68, opened 2025-08-15
chore: Address TODOs

iverc:
## Description

This PR address multiple TODOs across the codebase, including missing tests, txBuilder transaction evaluation, signed policy command validations, and duplicate payload id prevention


sdk/tx.go
```diff
@@ -94,6 +104,9 @@ func (b *TxBuilder) initTx(ctx context.Context, signer TxSigner, msgs ...sdk.Msg
 	txBuilder.SetGasLimit(b.gasLimit)
 	feeAmt := sdk.NewCoins(sdk.NewInt64Coin(b.feeTokenDenom, b.feeAmt))
 	txBuilder.SetFeeAmount(feeAmt)
+	if b.feeGranter != nil {
```
Lodek, line 107: nice catch

x/acp/signed_policy_cmd/payload_id.go
```diff
@@ -0,0 +1,38 @@
...
+		if bz, err := payload.Cmd.Marshal(); err == nil {
+			utils.WriteBytes(hasher, bz)
+		} else {
+			utils.WriteBytes(hasher, []byte(fmt.Sprintf("%T", payload.Cmd)))
+		}
+	} else {
+		utils.WriteBytes(hasher, nil)
+	}
+
+	return base32.StdEncoding.EncodeToString(hasher.Sum(nil))
```
Lodek: I think we should focus on hashing the raw JWS instead, since that's effectively what we are trying to prevent from being replaying (`ie `MsgSignedPolicyCmd.payload`)

x/acp/keeper/keeper.go
```diff
@@ -146,6 +147,41 @@ func (k *Keeper) getPolicyCmdHandler(ctx sdk.Context) *policy_cmd.Handler {
...
+	key := []byte(id)
+	if pref.Has(key) {
+		return fmt.Errorf("signed policy cmd already processed")
+	}
+	var bz [8]byte
+	binary.BigEndian.PutUint64(bz[:], expireHeight)
+	pref.Set(key, bz[:])
+	return nil
+}
+
```
Lodek: request: All of the ACP module uses raccoon to manipulate the KV store, I'd prefer to follow that pattern. If you'd prefer, I can do the refactor so you don't have to waste time figuring out raccoon.

Lodek (commented):
Nice job! Thanks for the work Ivan.
Two primary pointers are using raccoon in the acp module and using the hash of the raw JWS as ID instead.
Let me know if you'd prefer that I do the raccoon part, it'd be no problem.

Lodek (approved):
Nice!

Approving as this all looks good, but there's a small detail regarding the pruning which will only happen if the same jws is used later on. However, since the jws tokens should be unique, it won't be pruned under the normal flow. 
I will address this at a later PR though, no reason to block this any longer :+1: .

PR #69, opened 2025-08-18
feat: User subscriptions

iverc:
## Description

This PR adds support for managing user subscriptions, related developer configurations, and adds multiple helpers to automate developer lockup/credit management across epochs. 

Supported txs:
- `rpc CreateDeveloper(MsgCreateDeveloper)` - creates a developer record.
- `rpc UpdateDeveloper(MsgUpdateDeveloper)` - updates a developer record.
- `rpc RemoveDeveloper(MsgRemoveDeveloper)` - removes a developer record.
- `rpc AddUserSubscription(MsgAddUserSubscription)` - adds user subscription to a developer.
- `rpc UpdateUserSubscription(MsgUpdateUserSubscription)` - updates user subscription of a developer.
- `rpc RemoveUserSubscription(MsgRemoveUserSubscription)` - removes user subscription from a developer.

Supported queries:
- `rpc Developers(DevelopersRequest)` - lists all registered developers.
- `rpc UserSubscriptions(UserSubscriptionsRequest)` - lists all user subscriptions for a specified developer.


Lodek (commented):
I've been thinking about this for a while now and after our chat last week about the Trust API, I want to suggest another angle for the user authentication / user identifiers.

TLDR the tier module understands users as sourcehub accounts but IMO they should be dids instead.

Relevant bit:
```proto
// UserSubscription represents a user subscribed to a developer.
message UserSubscription {
  // developer is the address of the developer who registered this user
  string developer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
  // user is the address of the registered user
  string user = 2 [(cosmos_proto.scalar) = "cosmos.AddressString"];
  // SNIP
}
```

In the current e2e flow for Source's stack, users will create key pairs (either using Defra or locally), generate a JWT token and send HTTP request to Defra with this token. Defra will use the JWT token and broadcast a `MsgBearerPolicyCmd` to perform ACP operations. Right now, the final system users are `did`s, and I think we should keep that going to ensure a transparent operation of the system. Just as the ACP system supports DID actors, I think the same should apply to the tier module in order to track credit usage.

Happy to have a chat about this to brainstorm and spot some pitfalls.

proto/sourcehub/tier/v1beta1/tx.proto
```diff
@@ -119,16 +138,99 @@ message MsgCancelUnlocking {
...
 // MsgCancelUnlockingResponse defines the response structure for executing a MsgCancelUnlocking message.
 message MsgCancelUnlockingResponse {}
+
+// MsgCreateDeveloper is the Msg/CreateDeveloper request type.
+message MsgCreateDeveloper {
+  option (cosmos.msg.v1.signer) = "developer";
+  // developer is the address of the developer being created.
+  string developer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // auto_lock_enabled indicates whether the module should auto-lock to ensure enough credits.
+  bool auto_lock_enabled = 2;
```
Lodek, line 160: todo: add more context about the auto lock flag. End users won't be sure what this means

tests/integration/tier/test_helpers.go
Lodek, line 1: suggestion: It'd be better to encapsulate the methods in this file in a struct. That way forgetting to reset the indices doesn't risk causing a panic or if we ever decide to run tests in parallel we won't need to worry about race conditions

tests/integration/tier/tier_test.go
```diff
@@ -0,0 +1,501 @@
...
+		resp, err := suite.msgServer.UpdateDeveloper(suite.ctx, updateMsg)
+		require.NoError(t, err)
+		require.NotNil(t, resp)
+
+		dev := suite.keeper.GetDeveloper(suite.ctx, developer)
+		require.NotNil(t, dev)
+		require.True(t, dev.AutoLockEnabled)
+	})
+
+	suite.T().Run("Update non-existent developer creates new one", func(t *testing.T) {
```
Lodek: question: Do we need create / update methods if they effectively do the same thing? We could use a single `SetDeveloper` method instead.

tests/integration/tier/tier_test.go
```diff
@@ -0,0 +1,501 @@
...
+			AutoLockEnabled: true,
+		}
+
+		err := msg.ValidateBasic()
+		require.Error(t, err)
+	})
+}
+
+func (suite *TierIntegrationTestSuite) TestUpdateDeveloper() {
+	suite.T().Run("Update existing developer", func(t *testing.T) {
```
Lodek: suggestion: the nested suites are a bit trippy. IMO creating a dedicated struct for `TestUpdateDeveloperSuite` would be cleaner

tests/integration/tier/tier_test.go
```diff
@@ -0,0 +1,501 @@
...
+		}
+
+		_, err := suite.msgServer.RemoveUserSubscription(suite.ctx, removeMsg)
+		require.Error(t, err)
+		require.ErrorContains(t, err, "not subscribed")
+	})
+}
+
+func (suite *TierIntegrationTestSuite) TestRemoveDeveloperWithSubscriptions() {
+	suite.T().Run("Remove developer should remove all subscriptions", func(t *testing.T) {
```
Lodek: shouldn't the subscriptions stay active until the end of the epoch, even if the developer gets removed?
there's a potential for abused here, where developers could delete themselves and readd themselves, wiping the user subscriptions after getting paid

x/tier/keeper/keeper.go
```diff
@@ -497,3 +508,388 @@ func (k *Keeper) CancelUnlocking(ctx context.Context, delAddr sdk.AccAddress, va
...
+
+	k.SetDeveloper(ctx, developerAddr, developer)
+
+	sdk.UnwrapSDKContext(ctx).EventManager().EmitEvent(
+		sdk.NewEvent(
+			types.EventTypeDeveloperCreated,
+			sdk.NewAttribute(types.AttributeKeyDeveloper, developerAddr.String()),
+			sdk.NewAttribute(types.AttributeKeyAutoLockEnabled, fmt.Sprintf("%t", autoLockEnabled)),
+		),
+	)
```
Lodek, line 547: suggestion: it'd easier for consumers to consume typed events

x/tier/keeper/keeper.go
```diff
@@ -497,3 +508,388 @@ func (k *Keeper) CancelUnlocking(ctx context.Context, delAddr sdk.AccAddress, va
...
+		)
+	}()
+
+	existingSub := k.GetUserSubscription(ctx, developerAddr, userDid)
+	if existingSub != nil {
+		return types.ErrInvalidAddress.Wrapf("user %s is already subscribed to developer %s", userDid, developerAddr.String())
+	}
+
+	if amount != nil {
+		if amount.Denom != appparams.MicroCreditDenom {
```
Lodek: suggestion: instead of using coins we could use an unit for credits in the subscription directly.

proto/sourcehub/tier/v1beta1/subscription.proto
```diff
@@ -0,0 +1,56 @@
...
+}
+
+// UserSubscription represents a user subscribed to a developer.
+message UserSubscription {
+  // developer is the address of the developer who registered this user
+  string developer = 1 [(cosmos_proto.scalar) = "cosmos.AddressString"];
+  // user_did is the DID of the registered user
+  string user_did = 2;
+  // credit_amount is the credit amount allocated to this user
+  cosmos.base.v1beta1.Coin credit_amount = 3 [
```
Lodek: suggestion: change from coint to uint64, since we only support credits

Lodek: suggestion: change from coint to uint64, since we only support credits

x/tier/keeper/msg_server.go
Lodek, line 1: suggestion: Move the metrics which collect duration measurements to this file. We care about how long methods take to execute from the perspective of the *end* user. The internal methods are a consequence. It's far easier for us to monitor if there's a spike in latency to one specific operation with a higher level view. We can use other techniques like tracing to troubleshoot.

For context:
> Does this alert definitely indicate that users are being negatively affected?
https://sre.google/sre-book/monitoring-distributed-systems/
It's better if an issue we spot is tied to something the users may experience, internal methods can be used in other contexts like batch jobs (eg end blockers) which can skew measurements

x/tier/keeper/keeper.go
```diff
@@ -497,3 +508,388 @@ func (k *Keeper) CancelUnlocking(ctx context.Context, delAddr sdk.AccAddress, va
...
+		delta := newAmount.Sub(oldAmount)
+		if err := k.updateDeveloperTotalGranted(ctx, developerAddr, delta, true); err != nil {
+			return err
+		}
+	} else if oldAmount.GT(newAmount) {
+		delta := oldAmount.Sub(newAmount)
+		if err := k.updateDeveloperTotalGranted(ctx, developerAddr, delta, false); err != nil {
+			return err
+		}
+	}
```
Lodek: should we support this? If me as an user paid for my subscription for a period, I wouldn't want the developer to deduct my total amount of credits after it's been paid for.

x/tier/keeper/msg_remove_user_subscription_test.go
```diff
@@ -0,0 +1,253 @@
...
+	err = k.AddUserSubscription(sdkCtx, developerAddr, userDid, &amount, 30)
+	require.NoError(t, err)
+
+	testCases := []struct {
+		name      string
+		input     *types.MsgRemoveUserSubscription
+		expErr    bool
+		expErrMsg string
+	}{
+		{
```
Lodek, line 49: question: does the subscription remains active until the period expires?

tests/integration/tier/tier_test.go
iverc: Yes, I believe makes sense to have two separate methods, especially in terms of observability and separation of emitted events.
However, I think we could update `UpdateDeveloper` so that it does not create developer when trying to update non-existent developer to make sure that users are more intentional about using create vs update. 
Thanks for the feedback @Lodek

tests/integration/tier/tier_test.go
iverc: @Lodek subscriptions should stay active until the end of the current subscription period (which is not always the same as the epoch duration), and that's how it currently works. The actual subscription record is removed right away along with the developer record, but the associated feegrant allowance is set to expire at the current period reset time.

Updated the comments to make it clear what is actually happening. Thanks!

x/tier/keeper/msg_server.go
iverc: Thanks for the suggestion, I agree that it would be beneficial to move metrics to msg_server methods. I'll make this change as part of the upcoming PR that has various other tier module fixes/improvements.

x/tier/keeper/keeper.go
iverc: @Lodek we are not deducting the total credit amount here, the `delta` is used to accurately update the total granted amount for that developer after updating the subscription.

x/tier/keeper/msg_remove_user_subscription_test.go
iverc: yes

x/tier/keeper/keeper.go
iverc: @Lodek 💯
I'll make this change as part of the upcoming PR that has various other fixes/improvements, including events refactoring.

Lodek (approved):
lgtm, thanks for addressing the suggestions

PR #72, opened 2025-09-03
feat: IBC ACP (v2)

iverc:
## Description

This PR improves the ICA support implemented in #66 by adding ICA host middleware for connection tracking and enhancing the process of issuing DIDs for interchain accounts when creating/editing policies, verifying access requests, and sending direct policy commands.


x/acp/keeper/ica_connection.go
```diff
@@ -0,0 +1,138 @@
...
+		return
+	}
+
+	// Check if connection already exists to avoid duplicates
+	if _, exists := k.GetICAConnection(ctx, icaAddr); exists {
+		return
+	}
+
+	// Extract controller address from port ID (icacontroller-{address})
+	if !strings.HasPrefix(controllerPortID, "icacontroller-") {
```
Lodek: nit: Use the ica module constant instead: 
https://github.com/cosmos/ibc-go/blob/v10.3.0/modules/apps/27-interchain-accounts/types/keys.go#L15

app/ica_host_middleware.go
```diff
@@ -0,0 +1,110 @@
...
+	counterparty channeltypes.Counterparty,
+	counterpartyVersion string,
+) (string, error) {
+	version, err := m.next.OnChanOpenTry(ctx, order, connectionHops, portID, channelID, counterparty, counterpartyVersion)
+	if err != nil {
+		return "", err
+	}
+
+	// Only process ICA host channels during DeliverTx to avoid duplicate processing
+	if portID == "icahost" && len(connectionHops) > 0 && !ctx.IsCheckTx() {
```
Lodek: nit: compare against the constant in the ica mod instead: https://github.com/cosmos/ibc-go/blob/v10.3.0/modules/apps/27-interchain-accounts/types/keys.go#L12

x/acp/keeper/ica_connection.go
```diff
@@ -0,0 +1,138 @@
...
+
+	"github.com/sourcenetwork/sourcehub/x/acp/types"
+)
+
+// icaConnectionStore returns a prefix store for ICA connections.
+func (k *Keeper) icaConnectionStore(ctx sdk.Context) prefix.Store {
+	storeAdapter := runtime.KVStoreAdapter(k.storeService.OpenKVStore(ctx))
+	return prefix.NewStore(storeAdapter, []byte(types.ICAConnectionKeyPrefix))
+}
+
```
Lodek, line 22: in general, if we decide to keep this ica / ibc logic in the ACP module, I would much prefer if it done using a repository and Raccoon's ObjStore to maintain regularity within the module

x/acp/keeper/ica_connection.go
```diff
@@ -0,0 +1,138 @@
...
+	for ; iterator.Valid(); iterator.Next() {
+		var connection types.ICAConnection
+		k.cdc.MustUnmarshal(iterator.Value(), &connection)
+		icaAddress := string(iterator.Key())
+		cb(icaAddress, connection)
+	}
+}
+
+// HandleICAChannelOpen processes ICA channel opening and stores connection information.
+func (k *Keeper) HandleICAChannelOpen(
```
Lodek, line 89: todo: add logging to the empty returns. In general these should be edge cases (in my understanding) but we shouldn't fail silently if there's a chance this may happen.

x/acp/keeper/msg_server_direct_policy_cmd.go
```diff
@@ -5,31 +5,16 @@ import (
...
-		return nil, fmt.Errorf("DirectPolicyCmd: %v: %w", err, types.NewErrInvalidAccAddrErr(err, msg.Creator))
-	}
-
-	acc := k.accountKeeper.GetAccount(ctx, addr)
-	if acc == nil {
-		return nil, fmt.Errorf("DirectPolicyCmd: %w", types.NewAccNotFoundErr(msg.Creator))
-	}
-
-	actorID, err := did.IssueDID(acc)
+	actorID, err := k.IssueDIDFromAccountAddr(goCtx, msg.Creator)
```
Lodek, line 15: praise: thanks for fixing this, was an overlook on my part

Lodek (commented):
Nice! Good job Ivan. Apologies for the delayed review but took me a while to get onboarded with IBC / ICA.

In general, lgtm but I have a discussion point / suggestion. Should the ICA Connection logic be kept in the ACP module? It's my understanding shinzo will also use the bulletin module in the future. Perhaps it would be better and more generic to have another module which deals with the ICA stuff, seeing as there is nothing about this specific to ACP.

x/ica/keeper/ica_connection.go
```diff
@@ -0,0 +1,147 @@
...
+	if controllerChainID == "" {
+		ctx.Logger().Error("Controller chain ID is empty", "client_id", conn.ClientId)
+		return
+	}
+
+	controllerAddr := strings.TrimPrefix(controllerPortID, icatypes.ControllerPortPrefix)
+	if err := k.SetICAConnection(ctx, icaAddr, controllerAddr, controllerChainID, connectionID); err != nil {
+		ctx.Logger().Error("Failed to store ICA connection", "error", err, "ica_address", icaAddr)
+		return
+	}
```
Lodek, line 151: suggestion: I think it would be better to have the keeper return the error and log it on the middleware, that way it's easier to test those error cases in the future if we decide to

x/ica/keeper/ica_connection_test.go
```diff
@@ -0,0 +1,210 @@
...
+		require.Equal(t, icaAddress, iterIcaAddress)
+		require.Equal(t, icaAddress, connection.IcaAddress)
+		require.Equal(t, controllerAddress, connection.ControllerAddress)
+		require.Equal(t, controllerChainID, connection.ControllerChainId)
+		require.Equal(t, connectionID, connection.ConnectionId)
+		count++
+	})
+
+	require.Equal(t, 1, count)
+}
```
Lodek, line 133: suggestion: this test is redundant with the previous one, we can remove it.

Lodek (approved):

PR #76, opened 2025-09-26
feat: Custom Feegrant 

iverc:
## Description

This PR extends the standard feegrant module functionality to support did-based allowances, adds support for custom extension options, and introduces new faucet endpoint to grant did-based allowances.

New messages:
 - `MsgGrantDIDAllowance` - grants a fee allowance from granter to a DID.
 - `MsgRevokeDIDAllowance` - revokes a fee allowance between granter and DID.
 - `MsgPruneDIDAllowances` - prunes expired DID fee allowances.

  New queries:
  - `DIDAllowance` - queries allowance for a specific granter-DID pair.
  - `DIDAllowancesByGranter` - queries all the DID grants given by an address.


proto/sourcehub/faucet/faucet.proto
```diff
@@ -90,4 +90,31 @@ message GrantAllowanceResponse {
...
+// GrantDIDAllowanceRequest represents a request to grant fee allowance to a DID from the faucet.
+message GrantDIDAllowanceRequest {
+  string did = 1;
+  // Optional amount limit for the allowance (in uopen). If not provided, defaults to 10000000000 (10,000 OPEN)
+  cosmos.base.v1beta1.Coin amount_limit = 2 [
+    (gogoproto.nullable) = false,
+    (amino.dont_omitempty) = true
+  ];
+  // Optional expiration time in unix timestamp. If not provided, defaults to 30 days from now
+  int64 expiration = 3;
```
Lodek: question: is there a reason we can't use timestamps here? They'd be more convenient to inspect

proto/sourcehub/feegrant/v1beta1/tx.proto
```diff
@@ -74,3 +83,64 @@ message MsgPruneAllowances {
...
+
+  // allowance can be any of basic, periodic, allowed fee allowance.
+  google.protobuf.Any allowance = 3 [(cosmos_proto.accepts_interface) = "sourcehub.feegrant.v1beta1.FeeAllowanceI"];
+}
+
+// MsgGrantDIDAllowanceResponse defines the Msg/GrantDIDAllowanceResponse response type.
+message MsgGrantDIDAllowanceResponse {}
+
+// MsgRevokeDIDAllowance removes any existing Allowance from Granter to a DID.
+message MsgRevokeDIDAllowance {
```
Lodek, line 107: question: how does the protocol prevent a malicious developer from grabbing an users money and revoking their credits?

app/ante/extension.go
```diff
@@ -0,0 +1,100 @@
...
+				if jwsOpt, ok := extOpt.GetCachedValue().(*antetypes.JWSExtensionOption); ok {
+					currentTime := ctx.BlockTime()
+
+					// Validate JWS format, signature, required claims, and timing
+					did, authorizedAccount, err := validateJWSExtension(ctx, jwsOpt.BearerToken, currentTime)
+					if err != nil {
+						return ctx, errorsmod.Wrapf(sdkerrors.ErrInvalidRequest, "bearer token validation failed: %v", err)
+					}
+
+					// Verify authorized account matches the first message creator/signer
```
Lodek: suggestion: I think we should assert ALL messages in the Tx are signed by the authorized account. If the Tx is signed by multiple accounts I would suggest rejecting the Tx

app/ante/extension.go
```diff
@@ -0,0 +1,100 @@
...
+					// Try GetCreator for modern messages
+					if msgWithCreator, ok := firstMsg.(interface{ GetCreator() string }); ok {
+						msgSigner = msgWithCreator.GetCreator()
+					} else if legacyMsg, ok := firstMsg.(sdk.LegacyMsg); ok {
+						// Fallback to GetSigners for legacy messages
+						signers := legacyMsg.GetSigners()
+						if len(signers) > 0 {
+							msgSigner = signers[0].String()
+						}
+					}
```
Lodek: for Legacy Messages or Amino signatures we MUST reject the Tx.
https://github.com/cosmos/cosmos-sdk/issues/23859#issuecomment-2695240890

testutil/jws.go
```diff
@@ -0,0 +1,91 @@
...
+type BearerToken struct {
+	// IssuerID is the Actor ID for the Token signer
+	IssuerID string `json:"iss,omitempty"`
+	// AuthorizedAccount is the SourceHub account address which is allowed to use this token
+	AuthorizedAccount string `json:"authorized_account,omitempty"`
+	// IssuedTime is the timestamp at which the token was generated
+	IssuedTime int64 `json:"iat,omitempty"`
+	// ExpirationTime is the timestamp at which the token will expire
+	ExpirationTime int64 `json:"exp,omitempty"`
+}
```
Lodek: question: Is there a reason to duplicate this type? It's already defined in [app/ante/extension_utils.go](https://github.com/sourcenetwork/sourcehub/pull/76/files#diff-e6ab5aa65db1fae0e04c250cc77fad4395919a77591efbc56b765ac5bc12ea7c)

app/ante/extension_utils.go
```diff
@@ -0,0 +1,210 @@
...
+	}
+}
+
+// CreateJWSHeader creates the standard JWS header for EdDSA JWT tokens
+func CreateJWSHeader() map[string]any {
+	return map[string]any{
+		"alg": "EdDSA",
+		"typ": "JWT",
+	}
+}
```
Lodek: i think this is duplicated in the test file and IMO should be kept in the test file. Better not to expose this in the public API.

tests/integration/ante/jws_test.go
```diff
@@ -0,0 +1,238 @@
...
+
+	// Create transaction factory with proper account info
+	txf := clienttx.Factory{}.
+		WithTxConfig(clientCtx.TxConfig).
+		WithChainID(clientCtx.ChainID).
+		WithKeybase(clientCtx.Keyring).
+		WithSignMode(signing.SignMode_SIGN_MODE_DIRECT).
+		WithAccountRetriever(clientCtx.AccountRetriever).
+		WithAccountNumber(accNum).
+		WithSequence(seq)
```
Lodek, line 117: suggestion: We should support adding a Bearer Token in the sourcehub sdk package TxBuilder as well

app/ante/extension_utils.go
```diff
@@ -0,0 +1,210 @@
...
+	if secpKey, ok := pubKey.(secp.PublicKey); ok {
+		// https://www.rfc-editor.org/rfc/rfc8812
+		algs = []jwa.SignatureAlgorithm{jwa.ES256K}
+		pubKey = secpKey.ToECDSA()
+	} else {
+		algs, err = jwxjws.AlgorithmsForKey(pubKey)
+		if err != nil {
+			return BearerToken{}, fmt.Errorf("failed to retrieve algs for pub key: %v", err)
+		}
+	}
```
Lodek, line 69: one todo which I realized only after the original implementation, to be compliant with the JWS spec, we should assert that `alg` header in the JWS matches the alg derived from the did key. Non blocker atm, not a correctness issue either, just something we should address to be "more spec compliant".

tests/integration/ante/jws_test.go
```diff
@@ -0,0 +1,238 @@
...
+	"github.com/sourcenetwork/sourcehub/app/ante"
+	jwstypes "github.com/sourcenetwork/sourcehub/app/ante/types"
+	"github.com/sourcenetwork/sourcehub/testutil/network"
+	acptypes "github.com/sourcenetwork/sourcehub/x/acp/types"
+	"github.com/sourcenetwork/sourcehub/x/feegrant"
+)
+
+// TestJWSExtensionOptionWithDIDBasedFeegrant ensures that tx with JWS extension succeeds using the DID-based feegrant.
+func TestJWSExtensionOptionWithDIDBasedFeegrant(t *testing.T) {
+	net := network.NewWithOptions(t, network.NetworkOptions{EnableFaucet: true})
```
Lodek, line 34: IMO this is more of an e2e test, so it should live in the tests/e2e package

x/feegrant/keeper/keeper.go
```diff
@@ -359,3 +386,271 @@ func (k Keeper) RemoveExpiredAllowances(ctx context.Context, limit int32) error
...
+			return err
+		}
+		stop = cb(didGrant)
+	}
+
+	return nil
+}
+
+// GrantDIDAllowance creates a new DID-based grant.
+func (k Keeper) GrantDIDAllowance(
```
Lodek, line 445: Just one question for my own understanding. What happens if the granter runs out of fund? Does the error happen in the ante handler?

Lodek (approved):

PR #81, opened 2025-11-07
feat: Tier improvements

iverc:
## Description

This PR adds multiple improvements and fixes to the tier module, including auto-lock message for lockups with automatic validator selection and developer did extraction in subscription operations (e.g. add/update/remove).


x/tier/types/expected_keepers.go
```diff
@@ -87,3 +87,13 @@ type ParamSubspace interface {
...
 }
+
+// AcpKeeper defines the expected interface for the ACP module.
+type AcpKeeper interface {
+	// GetActorDID returns the actor DID, checking context first, then falling back to issuing from account address
+	GetActorDID(ctx sdk.Context, accountAddr string) (string, error)
+	// IssueDIDFromAccountAddr creates a DID from an account address
+	IssueDIDFromAccountAddr(ctx context.Context, accountAddr string) (string, error)
+	// GetAddressFromDID extracts and returns the bech32 address from a DID
+	GetAddressFromDID(ctx context.Context, did string) (sdk.AccAddress, error)
```
Lodek, line 98: thought: since, DIDs are used chain wide, perhaps the did functionality can be moved to the hub module

x/tier/keeper/keeper.go
```diff
@@ -650,6 +775,15 @@ func (k *Keeper) AddUserSubscription(
...
 
+	if extractedDID, ok := ctx.Value(appparams.ExtractedDIDContextKey).(string); ok && extractedDID != "" {
+		didAddr, err := k.GetAcpKeeper().GetAddressFromDID(sdk.UnwrapSDKContext(ctx), extractedDID)
+		if err != nil {
+			k.Logger().Error("Failed to convert DID to address", "did", extractedDID, "error", err)
+			return errorsmod.Wrap(err, "convert DID to address")
+		}
+		developerAddr = didAddr
+	}
+
```
Lodek, line 806: question: I understand we want this for TrustAPI, should we also support developer dids for the permissionless deployments?

x/tier/keeper/keeper.go
```diff
@@ -166,6 +174,123 @@ func (k *Keeper) CompleteUnlocking(ctx context.Context) (err error) {
...
+		validator, err := k.stakingKeeper.GetValidator(ctx, minLockup.valAddr)
+		if err == nil && validator.IsBonded() {
+			valAddr = minLockup.valAddr
+			err = k.Lock(ctx, delAddr, valAddr, amt)
+			if err != nil {
+				return nil, err
+			}
+			return valAddr, nil
+		}
+	}
```
Lodek: suggestion: there are two strategies for the `AutoLock` feature, this could be broken up into two funcs

x/tier/types/expected_keepers.go
iverc: @Lodek thats a good idea, I think it should be done in a separate PR since this change would affect multiple modules

Lodek (approved):

PR #82, opened 2025-11-07
fix: JWS token validation

iverc:
## Description

This PR fixes a bug with jws token validation, allowing to invalidate expired tokens while preventing expired tokens from being created. Authorized account validation / indexing is performed only if it is set.

### DIFF
--- a/x/hub/keeper/jws_token.go
+++ b/x/hub/keeper/jws_token.go
@@ -50,6 +50,8 @@ func (k *Keeper) SetJWSToken(ctx context.Context, record *types.JWSTokenRecord) 
 		if _, err := sdk.AccAddressFromBech32(record.AuthorizedAccount); err != nil {
 			return fmt.Errorf("invalid authorized account address: %w", err)
 		}
+	} else if !k.IsBearerAuthIgnored(ctx) {
+		return fmt.Errorf("authorized account is required when bearer auth is enabled")
 	}
 
 	bz, err := k.cdc.Marshal(record)
--- a/x/hub/keeper/jws_token.go
+++ b/x/hub/keeper/jws_token.go
@@ -45,17 +45,11 @@ func (k *Keeper) SetJWSToken(ctx context.Context, record *types.JWSTokenRecord) 
 		return fmt.Errorf("issuer DID cannot be empty")
 	}
 
-	if record.AuthorizedAccount == "" {
-		return fmt.Errorf("authorized account cannot be empty")
-	}
-
-	if _, err := sdk.AccAddressFromBech32(record.AuthorizedAccount); err != nil {
-		return fmt.Errorf("invalid authorized account address: %w", err)
-	}
-
-	sdkCtx := sdk.UnwrapSDKContext(ctx)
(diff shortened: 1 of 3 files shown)


x/hub/keeper/jws_token.go
```diff
@@ -45,17 +45,11 @@ func (k *Keeper) SetJWSToken(ctx context.Context, record *types.JWSTokenRecord)
...
-	}
-
-	sdkCtx := sdk.UnwrapSDKContext(ctx)
-	if !record.ExpiresAt.IsZero() && record.ExpiresAt.Before(sdkCtx.BlockTime()) {
-		return fmt.Errorf("expiration time cannot be in the past")
+	// Validate authorized account if provided
+	if record.AuthorizedAccount != "" {
+		if _, err := sdk.AccAddressFromBech32(record.AuthorizedAccount); err != nil {
+			return fmt.Errorf("invalid authorized account address: %w", err)
+		}
```
Lodek, line 52: As a fallback, we should only skip this check if `IgnoreBearerAuth` is set. i know we already check that in antee but probably a good idea to do this here as well

x/hub/keeper/jws_token.go
```diff
@@ -73,12 +67,14 @@ func (k *Keeper) SetJWSToken(ctx context.Context, record *types.JWSTokenRecord)
...
 	didStore.Set(didKeyWithoutPrefix, []byte{0x01}) // Just a marker, actual data is in primary store
 
-	// Store in account index
-	accountStore := k.jwsTokenByAccountStore(ctx)
-	accountKey := types.JWSTokenByAccountKey(record.AuthorizedAccount, record.TokenHash)
-	// Remove the prefix since we're already in the account prefix store
-	accountKeyWithoutPrefix := accountKey[len(types.JWSTokenByAccountKeyPrefix):]
-	accountStore.Set(accountKeyWithoutPrefix, []byte{0x01}) // Just a marker
+	// Store in account index only if authorized account is provided
+	if record.AuthorizedAccount != "" {
```
Lodek, line 73: same here

x/hub/keeper/jws_token.go
```diff
@@ -119,11 +115,13 @@ func (k *Keeper) DeleteJWSToken(ctx context.Context, tokenHash string) error {
 	didKeyWithoutPrefix := didKey[len(types.JWSTokenByDIDKeyPrefix):]
 	didStore.Delete(didKeyWithoutPrefix)
 
-	// Delete from account index
-	accountStore := k.jwsTokenByAccountStore(ctx)
-	accountKey := types.JWSTokenByAccountKey(record.AuthorizedAccount, tokenHash)
-	accountKeyWithoutPrefix := accountKey[len(types.JWSTokenByAccountKeyPrefix):]
-	accountStore.Delete(accountKeyWithoutPrefix)
+	// Delete from account index only if authorized account was provided
+	if record.AuthorizedAccount != "" {
```
Lodek, line 121: same here

Lodek (approved):

PR #84, opened 2025-11-10
fix: Bulletin keys

iverc:
## Description

This PR improves the bulletin keys logic to sanitize key parts and use `|` instead of `/` to prevent possible key collisions.


Lodek (approved):
🫡

PR #88, opened 2025-11-24
refactor: Update Policy DSL in SourceHub

Lodek:
This PR bumps the acp_core dependency which introduces a new policy marshaling scheme


tests/integration/ante/jws_test.go
```diff
@@ -151,26 +151,26 @@ func TestJWSExtensionOptionWithDIDBasedFeegrant(t *testing.T) {
 
 	// Create policy message with JWS extension
 	policyContent := `
-name: Bulletin Policy
 description: Base policy that defines permissions for bulletin namespaces
+name: Bulletin Policy
```
iverc, line 155: question: is there a reason why we define description before the name in all the updated policy definitions?

x/acp/keeper/abci_test.go
```diff
@@ -30,8 +30,9 @@ func TestEndBlocker(t *testing.T) {
 	ctx, err := utils.InjectPrincipal(ctx, "did:example:bob")
 	require.NoError(t, err)
 	resp, err := engine.CreatePolicy(ctx, &coretypes.CreatePolicyRequest{
-		Policy:      `name: test`,
-		MarshalType: coretypes.PolicyMarshalingType_SHORT_YAML,
+		Policy: `
+name: test
+`, MarshalType: coretypes.PolicyMarshalingType_YAML,
```
iverc: nit: lets move MarshalType to a new line

x/acp/keeper/actor_test.go
```diff
@@ -111,15 +111,15 @@ func TestEditPolicy_WithExtractedDID(t *testing.T) {
...
-        types:
-          - actor
-`,
-		MarshalType: coretypes.PolicyMarshalingType_SHORT_YAML,
+- name: file
+  relations:
+  - name: owner
+    types:
+    - actor
+`, MarshalType: coretypes.PolicyMarshalingType_YAML,
```
iverc: same here

x/acp/keeper/actor_test.go
```diff
@@ -133,15 +133,15 @@ resources:
...
```
iverc: same here

x/acp/keeper/actor_test.go
```diff
@@ -166,15 +166,15 @@ func TestDirectPolicyCmd_WithExtractedDID(t *testing.T) {
...
```
iverc: same here

tests/integration/acp/actions.go
```diff
@@ -50,15 +50,14 @@ func (a *EditPolicyAction) Run(ctx *TestCtx) *coretypes.Policy {
...
-		MarshalType: coretypes.PolicyMarshalingType_SHORT_YAML,
+		MarshalType: coretypes.PolicyMarshalingType_YAML,
 	}
 	response, err := ctx.Executor.EditPolicy(ctx, msg)
 	a.Response = response
 
 	AssertError(ctx, err, a.ExpectedErr)
 	if a.Expected != nil {
 		require.NotNil(ctx.T, response)
-		AssertValue(ctx, response.Record.Policy, a.Expected)
```
iverc, line 61: question: why do we remove this assertion?

iverc (commented):
LGTM, nice job, Bruno!

Just a few questions/minor suggestions.

Also, it looks like we have removed the SHORT_JSON policy type, but there are still some references to it in multiple files (e.g. policy_short.pulsar.go, cmd/ica_packet_gen/main.go, etc).

tests/integration/ante/jws_test.go
Lodek: Since there were a lot of policies across the projects, I wrote a tool to automate the migration. The tool uses go json marshaling, which marshals struct fields in alphabetic order

tests/integration/acp/actions.go
Lodek: there are some inconsistencies regarding empty lists vs nil values, between the return type of the function. We assert this anyway in the follwoing step (because we execute a query), so it is a bit redundant and it was being a pain

tests/integration/acp/suite/policy/edit_test.go
```diff
@@ -36,74 +36,83 @@ resources:
...
-							},
-						},
-					},
 					{
 						Name: "writer",
 					},
 				},
 				Permissions: []*coretypes.Permission{
 					{
-						Name:       "_can_manage_collaborator",
```
iverc, line 61: question: is there a  change in acp_core that affects how bulletin works (e.g. do we need to set the `_can_manage_collaborator` permission manually now)?

Lodek: no, these changes are internal. The policy used to expose the management rules as normal permissions but I've changed this representation to a dedicated field in the resource definition. So long as callers didn't rely on these permissions explicitly (which they shouldn't have), there will be no changes.

iverc (approved):

PR #93, opened 2026-01-22
Add artifact to create post event

JesseAbram:
Adds an artifact for event tracking to Create Post in bulletin see https://github.com/sourcenetwork/orbis-rs/issues/22 for more details on why this was needed


Lodek (approved):
Hope you came out the other side stronger after coding some Go  🫡

PR #101, opened 2026-03-26
fix: Codecs

iverc:
## Description

This PR fixes codecs issues and cleans up related code.


x/acp/types/codec.go
```diff
@@ -26,8 +25,6 @@ func RegisterInterfaces(registry cdctypes.InterfaceRegistry) {
 	registry.RegisterImplementations((*sdk.Msg)(nil),
 		&MsgEditPolicy{},
 	)
-	// this line is used by starport scaffolding # 3
```
Lodek, line 29: don't we need these lines if we scaffold through ignite?

iverc: I don't think we use it anymore

Lodek: i confirmed it, it's no longer used. They rely on ASTs now.

Lodek (approved):

PR #108, opened 2026-04-21
feat: Bulletin improvements

iverc:
## Description

This PR adds changes to the bulletin module to adhere to the Orbis requirements.

### DIFF
--- a/proto/sourcehub/bulletin/events.proto
+++ b/proto/sourcehub/bulletin/events.proto
@@ -31,6 +31,21 @@ message EventPostCreated {
   string artifact = 5;
 }
 
+// EventPostUpdated is emitted when an existing post's payload is updated.
+// The post_id is unchanged across updates.
+message EventPostUpdated {
+  // namespace_id is the namespace the post belongs to.
+  string namespace_id = 1;
+  // post_id is the unique identifier of the post (stable across updates).
+  string post_id = 2;
+  // updater_did is the DID of the actor that performed the update.
+  string updater_did = 3;
+  // payload is the base64-encoded new content of the post.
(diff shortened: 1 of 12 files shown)


JesseAbram (approved):
This looks great and is what I needed for orbis for now thank you, I may come back and restrict the update post function to be more restrictive in the future

JesseAbram:
https://github.com/sourcenetwork/sourcehub/pull/108#discussion_r3120740635

this btw is a problem for signing derivation which can be secured on the orbis side if this goes in (can also handle it on the bulletin side too) Will open an issue when/if this is merged 

It is worth it to note that I only need this for one "type" of post, so I can lock this update down to that data structure, or keep it open and add the extra security I need to orbis

JesseAbram:
> [#108 (comment)](https://github.com/sourcenetwork/sourcehub/pull/108#discussion_r3120740635)
> 
> this btw is a problem for signing derivation which can be secured on the orbis side if this goes in (can also handle it on the bulletin side too) Will open an issue when/if this is merged
> 
> It is worth it to note that I only need this for one "type" of post, so I can lock this update down to that data structure, or keep it open and add the extra security I need to orbis

actually since I bind the derivation to the policy id etc [here](https://github.com/sourcenetwork/orbis-rs/blob/develop/crates/crypto/src/bls12_381/sign.rs#L337)  and the impl [here](https://github.com/sourcenetwork/orbis-rs/blob/develop/bin/orbis-node/src/sign/coordinator.rs#L301-L305). This is not a problem as any manipulation would have to carry the policy_id with it, which would then guard the derivation signs

PR #120, opened 2026-06-17
orbis module extras

JesseAbram:
This PR adds a reporting system to orbis so nodes can report on each other with a threshold signature and node demerit points are assigned and tracked based on that. 

It also adds a way for a node to delete a pending ring


testutil/network/network.go
```diff
@@ -99,6 +99,10 @@ func DefaultConfigWithOptions(options NetworkOptions) network.Config {
 	if err != nil {
 		panic(err)
 	}
+	// These integration tests only use validator 0. Keeping the default network
+	// single-validator also avoids a CometBFT v0.38.21 shutdown race where a
+	// peer-reactor goroutine can access the block store after it has been closed.
+	cfg.NumValidators = 1
```
JesseAbram, line 105: double check this should not effect anything prod, just on the testing this should fix the flakey faucet tests

x/orbis/keeper/msg_server.go
```diff
@@ -126,7 +134,15 @@ func (k *Keeper) FinalizeRing(goCtx context.Context, msg *types.MsgFinalizeRing)
...
-			return nil, types.ErrRingPkConflict
+			if err := ctx.EventManager().EmitTypedEvent(&types.EventRingDeleted{
+				RingId: ring.Id,
+				Reason: "ring_pk_conflict",
+			}); err != nil {
+				return nil, err
+			}
+			return &types.MsgFinalizeRingResponse{
+				Outcome: types.FinalizeRingOutcome_CONFLICT_DELETED,
+			}, nil
```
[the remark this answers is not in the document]
JesseAbram: yes this was by design as one policy can be used on many rings and they get whitelisted on the node, this means that deleting one can brick multiple rings accidentally, huge footgun

x/orbis/keeper/validation.go
```diff
@@ -32,6 +32,9 @@ func validateRing(ring *types.Ring) error {
 	if err := validateUpgradeInfo(&ring.UpgradeInfo); err != nil {
 		return err
 	}
+	if err := types.ValidateDemeritConfig(ring.DemeritConfig); err != nil {
+		return err
+	}
```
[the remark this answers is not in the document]
JesseAbram: this was moved out in an early PR as it broke integration tests without actually adding security

iverc (commented):
I think we need to add genesis fields for demerits (genesis.go). a round-trip test for import/export would be nice too.

x/orbis/keeper/report.go
```diff
@@ -0,0 +1,549 @@
...
+	}
+
+	message, reportID, err := reportEnvelopeCanonicalMessageAndID(report)
+	if err != nil {
+		return nil, err
+	}
+	if claimedReportID != reportID {
+		return nil, errorsmod.Wrap(types.ErrInvalidReport, "report_id does not match canonical report bytes")
+	}
+	if k.HasAcceptedReport(goCtx, reportID) {
```
iverc, line 83: nit: I think it could be more efficient to check/dedupe it per ring/node/type/time so one outage only counts once

x/orbis/keeper/report.go
```diff
@@ -0,0 +1,549 @@
...
+	accusedCommittee, err := reportCommitteeForScope(ring, payload.accusedCommitteeScope)
+	if err != nil {
+		return err
+	}
+	signingCommittee, err := reportCommitteeForScope(ring, payload.signingCommitteeScope)
+	if err != nil {
+		return err
+	}
+
+	if signingCommittee.threshold < 2 {
```
iverc, line 307: nit: please add more tests around this logic

x/orbis/keeper/store.go
```diff
@@ -39,6 +40,85 @@ func (k *Keeper) DeleteRing(ctx context.Context, ringID string) {
 	store.Delete([]byte(ringID))
 }
 
+func (k *Keeper) SetAcceptedReport(ctx context.Context, reportID string) {
```
iverc: I think we should expire these in EndBlock after 120s

x/orbis/keeper/msg_server.go
```diff
@@ -161,7 +179,48 @@ func (k *Keeper) FinalizeRing(goCtx context.Context, msg *types.MsgFinalizeRing)
...
+		if err != nil {
+			return nil, err
+		}
+		authorized = cancellerDID == ring.CreatorDid
+	}
+	if !authorized {
+		return nil, types.ErrInvalidRingCanceller
+	}
+
+	k.DeleteRing(goCtx, ring.Id)
```
iverc, line 215: rings are hard deleted but acp objects are not, should we archive instead?

iverc (changes requested):
Left a few comments, mostly nits, except for the genesis.go import/export one. LGTM otherwise 👍

x/orbis/keeper/msg_server.go
JesseAbram: So my concern is that an acp policy can be used in multiple rings, so if a ring was deleted and the policy archived it would brick other rings?

x/orbis/keeper/report.go
JesseAbram: good find, I like this idea, I chose to solve it on the orbis side mainly as orbis should drive a lot of the security here. This is done by having a session id, a session id is created on every request so only one report per session stop duplicate sessions. The idea is there is one reason a session could fail, that creates a report. This also allows an extra check on the orbis side like hey this session failed a report request should be coming in, which I like

iverc (approved):

PR #122, opened 2026-07-02
Add Backup nodes and more reports

JesseAbram:


proto/sourcehub/orbis/ring.proto
```diff
@@ -37,7 +48,7 @@ message Ring {
   string policy_id = 10;
   repeated RingConfirmation confirmations = 11;
   UpgradeInfo upgrade_info = 12 [(gogoproto.nullable) = false];
-  DemeritConfig demerit_config = 13 [(gogoproto.nullable) = false];
+  ReportingConfig reporting = 13 [(gogoproto.nullable) = false];
```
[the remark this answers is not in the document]
JesseAbram: nothing is live yet no backwards comaptability needed

x/orbis/keeper/report.go
```diff
@@ -235,43 +374,965 @@ func reportEnvelopeCanonicalBytes(report *types.ReportEnvelope) ([]byte, error)
...
+	}
+	chainID, err := decoder.readString("chain_id")
+	if err != nil {
+		return invalidCryptoResponseStatement{}, err
+	}
+	ringID, err := decoder.readString("ring_id")
+	if err != nil {
+		return invalidCryptoResponseStatement{}, err
+	}
+	ringPk, err := decoder.readString("ring_pk")
```
iverc: nit: ideally we'd want to have these defines as constants too

x/orbis/keeper/report.go
```diff
@@ -377,6 +1465,84 @@ func reportBlockUnixTime(ctx sdk.Context) (uint64, error) {
...
+	ring *types.Ring,
+	accusedNodeKey string,
+) (string, []string, bool) {
+	active := make(map[string]struct{}, len(ring.PeerNodeKeys))
+	for _, nodeKey := range ring.PeerNodeKeys {
+		active[nodeKey] = struct{}{}
+	}
+
+	for i, backupNodeKey := range ring.Reporting.BackupNodeKeys {
+		if backupNodeKey == accusedNodeKey {
```
iverc: it seems like when the code skips the accused as a candidate (`if backupNodeKey == accusedNodeKey { continue }`) but never removes it from BackupNodeKeys

x/orbis/keeper/report.go
iverc, line 1: thought: we added `invalid_crypto_response_demerits` and `unauthorized_request_demerits` proto fields right after `reset_interval_seconds`, but the writer emits them in a different order.

x/orbis/keeper/report_test.go
iverc, line 1: nit: it seems like the `unauthorized_request` path is not fully tested to assert its demerit amount

iverc (approved):

x/orbis/keeper/report.go
JesseAbram: ya tbh this check is fully redundant and gonna remove because an accused key has to be either the current or next peer node keys and is handled in reportCommitteeForScope function

PR #128, opened 2026-07-27
fix: authenticate trusted relay workers

iverc:
## Summary
- require bearer relays to use a governance-configured trusted fee granter
- prove that the trusted account authorized each relay worker through the existing on-chain fee grant
- bind every relayed message creator to the fee-paying worker
- reject invalidated JWS tokens instead of only recording their status
- provision a separate funded relay control account in the local genesis fixture

## Why
`ignore_bearer_auth` allowed a worker pool to preserve the user DID while signing with rotating workers, but it also allowed any direct submitter to self-assert an actor DID. The trusted fee-granter check keeps the scalable worker model while anchoring relay authority to chain governance and an actual fee grant.

The JWS invalidation store also returned success for already-invalid tokens, and the ante decorator ignored storage errors. This made invalidation ineffective for transaction authorization.

## Verification
- `go test ./...`
- focused ante and Hub tests
- fresh `scripts/genesis-setup.sh` run
- `sourcehubd genesis validate-genesis`

Stacked on #127; merge #125, #126, #127 first.


app/ante/extension_utils.go
```diff
@@ -198,3 +198,10 @@ func getExtractedDIDFromContext(ctx sdk.Context) string {
 	}
 	return ""
 }
+
+func getTrustedRelayFeeGranterFromContext(ctx sdk.Context) string {
+	if address, ok := ctx.Value(appparams.TrustedRelayFeeGranterContextKey).(string); ok {
+		return address
+	}
+	return ""
```
JesseAbram, line 206: just curious why return an empty string and not an error?

JesseAbram (approved):

app/ante/extension_utils.go
iverc: in that case it would return empty string for regular/direct txs, not ones that come through the trust api

PR #130, opened 2026-08-12
feat(orbis): scope trusted auth relays to rings

iverc:
## Summary
- store the canonical trusted authentication relay set on each ring at creation
- bind non-empty initial relay sets to ring IDs and include current relay state in reporting and reshare signatures
- allow ACP-authorized relay revocation without permitting post-creation trust expansion

This is the SourceHub contract required by sourcenetwork/orbis-rs#229. Existing rings decode with an empty relay set and continue to use their existing IDs.

## Tests
- `go test ./...`
- cross-language canonical hash fixtures updated for the new ring state


x/orbis/keeper/msg_server.go
```diff
@@ -522,6 +524,48 @@ func (k *Keeper) SetRingReportingByAcp(goCtx context.Context, msg *types.MsgSetR
 	return &types.MsgSetRingReportingByAcpResponse{}, nil
 }
 
+func (k *Keeper) RemoveRingTrustedAuthRelayByAcp(
```
JesseAbram, line 581: If there is a remove should probably also be an add?

x/orbis/types/keys.go
```diff
@@ -37,18 +37,28 @@ func KeyPrefix(p string) []byte {
 // GenerateRingID returns the stable ID for a ring's creation parameters.
 // peerNodeKeys are sorted before hashing so the ID is order-independent.
 func GenerateRingID(peerNodeKeys []string, threshold uint32, pssInterval uint64, policyID string, nonce immutable.Option[string], currentVersion uint64) string {
-	sorted := slices.Clone(peerNodeKeys)
-	if !slices.IsSorted(sorted) {
-		slices.Sort(sorted)
+	return GenerateRingIDWithTrustedAuthRelays(peerNodeKeys, threshold, pssInterval, policyID, nonce, currentVersion, nil)
```
JesseAbram: there is not backwards compatibility needed here just update GenerateRingID to have sortedPeerNodeKeys and update the call sites, no need to split into two functions

JesseAbram (commented):
looks good 2 things before approving 

* I want to add a lock (maybe a flag in a ring) that says this will never have a trusted auth relay, and that can be set at the start.  This is good for a user that feels uncomfortable with this feature
* I need to review the pairing PR with orbis

proto/sourcehub/orbis/ring.proto
```diff
@@ -49,6 +49,10 @@ message Ring {
   repeated RingConfirmation confirmations = 11;
   UpgradeInfo upgrade_info = 12 [(gogoproto.nullable) = false];
   ReportingConfig reporting = 13 [(gogoproto.nullable) = false];
+  // Ed25519 did:key issuers allowed to authenticate requests for another actor.
+  repeated string trusted_auth_relay_dids = 14;
+  // Set at creation; false permanently disables trusted authentication relays.
+  bool allow_trusted_auth_relays = 15;
```
JesseAbram, line 55: I don't think this is used yet, as I don't see an add trusted_auth_relay function, this would be applied there and stop the addition if it was set to false........I guess that is a fork, if you don't allow additions to the trusted_auth_relays and just removals then this is not needed, however in case of lost keys or rotation, I lean towards having this and the addition function

JesseAbram (commented):
just the one item to fix

JesseAbram (approved):

PR #143, opened 2026-09-10
  Bind the ring public key into threshold-signature verification + reject degenerate ring keys

JesseAbram:
Two related hardening changes to x/orbis/keeper threshold-signature verification, both from the recent MPC-signing security review:

  1. Public-key-augmented BLS — verify BLS threshold signatures against H(ringPK ‖ message) instead of H(message), using the IETF AUG_ ciphersuite. Breaking wire change; must ship in lockstep with orbis-rs 
     Add-public-key-to-bls-message-hash.
  2. Reject identity / degenerate ring public keys at FinalizeRing and in the decaf377 FROST verifier.

  1. Public-key augmentation (BLS)

  Problem. orbis-rs derives per-request signing keys multiplicatively: PK = d·xG, and a plain BLS signature is d·x·H(m). The derivation scalar d is a deterministic function of on-chain data (derivation path + policy/resource/permission),
  so it is public. Given a signature σ_a valid under derived key PK_a, anyone can compute σ_v = (d_v · d_a⁻¹) · σ_a, which verifies under any other derived key PK_v on the same ring. Per-derivation access control on signing is
  bypassable by anyone holding any derivation on the ring, including the root key used for reports and ring updates.

  Fix. Message augmentation: hash H(PK ‖ m). H(PK_a ‖ m) and H(PK_v ‖ m) are independent curve points with no scalar relation, so no signature can be scaled onto another key. This matches draft-irtf-cfrg-bls-signature §3.2.

  Changes here:
  - bls12381G2SignatureDST → BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_AUG_ (renamed bls12381G2SignatureAugDST).
  - Scheme identifier bls12_381_g1_pk_g2_sig_nul → bls12_381_g1_pk_g2_sig_aug_v1 (renamed ThresholdSignatureSchemeBLS12381G1PKG2SigAugV1). This is the discriminator verifyThresholdSignature dispatches on; it travels in MsgSubmitReport /
    MsgFinalizeRingReshareByThresholdSignature.
  - verifyBLS12381ThresholdSignature passes useHash=true, usePksAsAugs=true to blst.VerifyCompressed, so blst prepends the exact decodeHexBytes(ring.RingPk) bytes — byte-identical to what orbis-rs prepends (ark-bls12-381 0.4 uses
    ZCash/blst-compatible point serialization, so pk.serialize_compressed() on the Rust side equals the stored ring.RingPk).
  - Both consumers (report.go, msg_server.go reshare finalize) verify root-key signatures against ring.RingPk and need no change beyond the dispatch.
  
  FROST/decaf377 needs no augmentation — its Fiat–Shamir challenge already binds the derived public key.

  2. Reject identity / degenerate ring public keys

  Problem. decaf377 Schnorr verification z·G == R + c·Y is vacuous when Y (the public key) is the identity: pick any z, set R = z·G, publish (R, z). orbisfrost.Verify accepted this. BLS was already protected by blst's KeyValidate /
  pkValidate.

  Changes here:
  - verifyDecaf377FROSTThresholdSignature rejects the decaf377 identity element before calling orbisfrost.Verify.
  - New rejectIdentityRingPublicKey, length-dispatched (decaf377 identity check; BLS Uncompress + !KeyValidate()), called from FinalizeRing so a ring can never be finalized with a degenerate key. Undecodable / unknown-length keys are
    left to the scheme-specific verifiers (FinalizeRing predates an explicit scheme field and some tests use placeholder keys).

### DIFF
--- a/x/orbis/keeper/threshold_signature.go
+++ b/x/orbis/keeper/threshold_signature.go
@@ -14,14 +14,21 @@ import (
 )
 
 const (
-	ThresholdSignatureSchemeBLS12381G1PKG2SigNUL = "bls12_381_g1_pk_g2_sig_nul"
-	ThresholdSignatureSchemeDecaf377FROST        = "decaf377_frost"
+	// ThresholdSignatureSchemeBLS12381G1PKG2SigAugV1 is the public-key-augmented
+	// BLS scheme: the ring public key is prepended to the message before
+	// hash-to-curve (IETF "message augmentation"), so a signature cannot be
+	// scaled from one publicly-related derived key onto another. Must stay in
+	// lockstep with orbis-rs `crypto::THRESHOLD_SIGNATURE_SCHEME`. The
+	// pre-augmentation scheme string was "bls12_381_g1_pk_g2_sig_nul".
+	ThresholdSignatureSchemeBLS12381G1PKG2SigAugV1 = "bls12_381_g1_pk_g2_sig_aug_v1"
+	ThresholdSignatureSchemeDecaf377FROST          = "decaf377_frost"
 
-	bls12381PublicKeySize  = 48
-	bls12381SignatureSize  = 96
-	bls12381G2SignatureDST = "BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_NUL_"
-	decaf377PublicKeySize  = decaf377.ElementSize
-	decaf377SignatureSize  = decaf377.ElementSize + decaf377.ScalarSize
+	bls12381PublicKeySize = 48
+	bls12381SignatureSize = 96
+	// AUG_ ciphersuite DST (matches orbis-rs `bls12_381::sign::BLS_SIG_DOMAIN`).
+	bls12381G2SignatureAugDST = "BLS_SIG_BLS12381G2_XMD:SHA-256_SSWU_RO_AUG_"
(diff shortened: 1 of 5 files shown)


x/orbis/keeper/threshold_signature.go
```diff
@@ -92,6 +113,32 @@ func verifyDecaf377FROSTThresholdSignature(ringPK string, message []byte, signat
...
+// key that makes Schnorr verification vacuous. FinalizeRing predates an
+// explicit curve/scheme field and its tests use opaque placeholder keys, so
+// other encoding validation remains in the scheme-specific consumers.
+func rejectIdentityRingPublicKey(ringPK string) error {
+	publicKey, err := decodeHexBytes(ringPK)
+	if err != nil {
+		return nil
+	}
+
+	switch len(publicKey) {
```
iverc, line 126: do we need to handle a default case here?

iverc (approved):

x/orbis/keeper/threshold_signature.go
JesseAbram: No need we are just checking that the ring's public key isnt the group identity element. We can do that by checking the size to get sig type and then applying the right function. A more comprehensive check of a junk key would be caught with ErrInvalidThresholdSignature

